PR #22, opened 2021-01-12
Add equals and notEquals operators

### LINKED ISSUE #15: Add support for the equals operator
This operator supports all JSON types

Tests the template value of the path to determine if it is equal to the value specified in the rule.

- If the type of the value of equals does not match the type of the value at path, this evaluates to false (except for integer and float, which can be compared with one another). Otherwise, this behaves as expected for the given type.
- Evaluations on string types are case-insensitive.

``` javascript
{
    "path": "name",
    "equals": "MyVMResource" // Evaluates to `true` because the value of the path "name" is a string and case-insensitively matches the value in the rule.
}
```

### LINKED ISSUE #16: Add support for notEquals operator
This operator supports all JSON types

The logical inverse of equals. Evaluations on incompatible types results in true.

``` javascript
{
    "path": "properties.osProfile.adminPassword",
    "notEquals": "password" // Evaluates to `true` because the value of the path "properties.osProfile.adminPassword" is `null`, which does not match the value in the rule.
}
```

yane3628:
- Add support for equals operator (Closes #15)
- Add support for notEquals operator (Closes #16)
- Add built-in rule that recommends not to use custom RBAC roles


src/Analyzer.JsonRuleEngine.UnitTests/EqualsOperatorTests.cs
```diff
@@ -0,0 +1,72 @@
...
+            Assert.IsTrue(equalsOperator.EvaluateExpression(jToken));
+        }
+
+        [DataTestMethod]
+        [DataRow("value", "value2", DisplayName = "String values are not equal")]
+        [DataRow(true, false, DisplayName = "Boolean values are not equal")]
+        [DataRow(1, 2, DisplayName = "Integer values are not equal")]
+        [DataRow(0.1, 0.2, DisplayName = "Float values are not equal")]
+        [DynamicData(nameof(TestArrays), DynamicDataSourceType.Method, DynamicDataDisplayName = "GetArrayValuesDynamicDataDisplayName")]
+        [DataRow(@"{""property"": ""value11""}", @"{""property2"": ""value12""}", DisplayName = "Json values are not equal")]
```
JohnathonMohr, line 37: Can there be similar tests for when `"Equals": value` is true?

src/Analyzer.JsonRuleEngine.UnitTests/EqualsOperatorTests.cs
```diff
@@ -0,0 +1,72 @@
...
+        public void EvaluateExpression_PropertyIsNotEqual_EqualsExpressionIsFalse(object expectedValue, object actualValue)
```
JohnathonMohr, line 38: Also add tests for comparing values of different types

src/Analyzer.JsonRuleEngine.UnitTests/EqualsOperatorTests.cs
```diff
@@ -0,0 +1,72 @@
...
+using System.Collections.Generic;
+using System.Reflection;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.Templates.Analyzer.JsonRuleEngine.UnitTests
+{
+    [TestClass]
+    public class EqualsOperatorTests
```
JohnathonMohr, line 13: Also add tests for NotEquals

JohnathonMohr (changes requested):

src/Analyzer.JsonRuleEngine.UnitTests/EqualsOperatorTests.cs
yane3628: This is the test on line 22

JohnathonMohr: Oh, yes, the slight difference is because they don't need to have different values for the positive test.  Thanks.

src/Analyzer.JsonRuleEngine.UnitTests/EqualsOperatorTests.cs
yane3628: What is the expectation here? Would it throw an error or simply return that it is not equal? Right now it does the later, and I can add another DataRow.

JohnathonMohr (approved):
I think this is a good start.  We should add a new issue to improve the Equals operator too, so that:
- not everything is based on string comparisons
- ensure comparing objects and arrays is not case-sensitive

PR #24, opened 2021-01-13
Enforcing path to be specified in Value Operators.

### LINKED ISSUE #23: Enforce path to be present in Value Operators
The [documentation](https://github.com/Azure/template-analyzer/blob/main/docs/authoring-json-rules.md#value-operators) explains that `path` must have a value in all Value Operators, but it currently isn't checked or enforced in code.  The Analyzer should not run rules where `path` is missing but should be present.

JohnathonMohr:
- Ensures that `path` has a value in all Value Operators before running rules. (Fixes #23)
- Refactored resource type filtering to improve testability.
- Improved those tests.


src/Analyzer.JsonRuleEngine.UnitTests/LeafExpressionTests.cs
```diff
@@ -49,29 +43,37 @@ public void Evaluate_ValidScope_ReturnsResultsOfOperatorEvaluation(string resour
             };
 
             // JObject to evaluate, in this test, this is a subset of an ARM template
-            var jsonToEvaluate = JObject.Parse(jtoken);
+            var jsonToEvaluate = JObject.Parse("{ \"property\": \"value\" }");
 
-            // Setting up the Mock JsonPathResolver to return the expected values when JToken and Resolve are called
+            // Setting up the Mock JsonPathResolvers to return the expected values when JToken and Resolve are called
             var mockJsonPathResolver = new Mock<IJsonPathResolver>();
+            var mockResourcesResolved = new Mock<IJsonPathResolver>();
```
yane3628, line 50: *mockResourceResolver

yane3628 (approved):

src/Analyzer.JsonRuleEngine.UnitTests/LeafExpressionTests.cs
JohnathonMohr: I chose the name `mockResourcesResolved` because it's not a "resource resolver", it's a mock of a resolved resource.  Is there a better name I can use to clarify this?

PR #26, opened 2021-01-15
Add code for CLI

varunsh-msft:


yane3628 (approved):

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -0,0 +1,139 @@
...
+
+            analyzeTemplateCommand.AddOption(parameterOption);
+
+            analyzeTemplateCommand.AddOption(SetupOutputFileOption());
+
+            analyzeTemplateCommand.Handler = CommandHandler.Create<FileInfo, FileInfo>((templateFilePath, parametersFilePath) =>
+            {
+                try
+                {
+                    Core.TemplateAnalyzer templateAnalyzer = new Core.TemplateAnalyzer(File.ReadAllText(templateFilePath.FullName), parametersFilePath == null ? null : File.ReadAllText(parametersFilePath.FullName));
```
JohnathonMohr, line 78: For testing purposes, instead of directly creating a `TemplateAnalyzer` here, pass in a builder to the constructor of this class that can create the `TemplateAnalyzer` with a template and parameters.  For now, we can keep the builder simple.  Something like this should be fine:
``` C#
// Constructor
public CommandLineParser(Func<string, string, TemplateAnalyzer> analyzerBuilder) { ... }

// Creating CommandLineParser
new CommandLineParser((template, parameters) => new TemplateAnalyzer(template, parameters));
```
And replace here with:
``` suggestion
... = analyzerBuilder(File.ReadAllText(...), ...);
```

Then in the tests, pass in a builder that will create a mock.

JohnathonMohr (changes requested):
Please add tests for this.  See comment below for suggestions.

Update: Tests will be added later.  Issue #28 created to track.

JohnathonMohr (approved):

src/Analyzer.Cli/CommandLineParser.cs
JohnathonMohr: Tagging issue #28.

Issue #30, opened 2021-01-20
Move JsonPathResolver.cs to Utilities project


JohnathonMohr:
Relocate the class to Utilities.  Potentially remove the interface it implements too, and in tests, mock the class directly instead of mocking the interface.

PR #32, opened 2021-01-22
Update Json rule engine namespaces

JohnathonMohr:
- Expanded namespaces within the Json rule engine
  - Namespace renamed from `Microsoft.Azure.Templates.Analyzer.JsonRuleEngine` to `Microsoft.Azure.Templates.Analyzer.RuleEngines.JsonEngine` so JsonRuleEngine isn't a namespace and class name
  - Kept the AssemblyName of the Json rule engine the same, since "....Analyzer.JsonRuleEngine.dll" seems like a better file name.
- Moved converter, operators, evaluations, and schemas into their own namespaces
- Moved `JsonPathResolver` to Utilities project, and its interface into Types (resolves issue #30)
- Updated test projects to mirror new namespace name

### DIFF
--- a/src/Analyzer.Utilities.UnitTests/JsonPathResolverTests.cs
+++ b/src/Analyzer.Utilities.UnitTests/JsonPathResolverTests.cs
@@ -7,7 +7,7 @@ using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
-namespace Microsoft.Azure.Templates.Analyzer.Utilities
+namespace Microsoft.Azure.Templates.Analyzer.Utilities.UnitTests
 {
     [TestClass]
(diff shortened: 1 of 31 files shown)


src/Analyzer.Utilities.UnitTests/JsonPathResolverTests.cs
```diff
@@ -7,7 +7,7 @@
 using Newtonsoft.Json.Linq;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
-namespace Microsoft.Azure.Templates.Analyzer.JsonRuleEngine.UnitTests
+namespace Microsoft.Azure.Templates.Analyzer.Utilities
```
yane3628, line 10: Namespace is missing *.UnitTests

yane3628 (approved):

PR #37, opened 2021-02-09
Add support for allOf Operator

### LINKED ISSUE #12: Add support for allOf operator
This operator is a logical and operator

``` javascript
{
    "anyOf": [
        {
            "path": "properties.osProfile.adminPassword",
            "hasValue": false // Evaluates to `false`
        },
        {
            "regex": "username" // Evaluates to `true`
        }
    ] // Evaluates to `false` because not all the expressions contained in "allOf" resulted in `true`
}
```

yane3628:
- Add support of allOf operator (Closes #12)
- Add tests for allOf
- Add allOf as keyword when parsing rules
- Validate allOf expression
- Add built-in rule that uses allOf
- Introduce Expression class for structured operators


src/Analyzer.JsonRuleEngine/Evaluation.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        /// <summary>
+        /// Gets all the results that evaluated to true.
+        /// </summary>
+        /// <returns>The results that evaluated to true.</returns>
+        public IEnumerable<JsonRuleResult> GetResultsEvaluatedTrue()
+        {
+            if (resultsEvaluatedTrue == null)
+            {
+                resultsEvaluatedTrue = Results.ToList().FindAll(r => r.Passed);
```
JohnathonMohr, line 43: There's a hidden bug here. :)
When this class is created, if `results` comes from the results of a function that uses `yield return` (most evaluations do), whichever of `GetResultsEvaluatedTrue` or `GetResultsEvaluatedFalse` is called first will iterate through the results and throw away some of them, and then the other function will not find anything.

You can test this by creating the results in the `Evaluate_TwoLeafExpressions_ExpectedResultIsReturned` test from `yield return` statements instead of any array.

I can explain in more detail offline if you want.

src/Analyzer.JsonRuleEngine/Evaluation.cs
```diff
@@ -0,0 +1,63 @@
...
+        /// </summary>
+        /// <param name="passed">Determines whether or not the rule for this evaluation passed.</param>
+        /// <param name="results"><see cref="IEnumerable"/> of results.</param>
+        public Evaluation(bool passed, IEnumerable<JsonRuleResult> results) => (this.Passed, this.Results) = (passed, results);
+
+        /// <summary>
+        /// Gets all the results that evaluated to true.
+        /// </summary>
+        /// <returns>The results that evaluated to true.</returns>
+        public IEnumerable<JsonRuleResult> GetResultsEvaluatedTrue()
```
JohnathonMohr, line 39: Can you change these functions to properties with only a get method instead (and remove "Get" from the name)?  It seems like a good fit since "Get" is the first word of the method.

src/Analyzer.JsonRuleEngine/Expressions/LeafExpression.cs
```diff
@@ -65,19 +65,30 @@ public override IEnumerable<JsonRuleResult> Evaluate(IJsonPathResolver jsonScope
...
                 var leafScope = scope?.Resolve(Path);
 
                 foreach (var propertyToEvaluate in leafScope)
                 {
-                    yield return new JsonRuleResult
+                    bool passed = Operator.EvaluateExpression(propertyToEvaluate.JToken);
+                    if (!passed && evaluationPassed)
+                    {
+                        evaluationPassed = false;
+                    }
```
JohnathonMohr, line 81:
``` suggestion
                    evaluationPassed &= passed;
```

src/Analyzer.JsonRuleEngine/Expressions/AllOfExpression.cs
```diff
@@ -0,0 +1,80 @@
...
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Azure.Templates.Analyzer.Types;
+
+namespace Microsoft.Azure.Templates.Analyzer.RuleEngines.JsonEngine.Expressions
+{
+    /// <summary>
+    /// Represents a all of expression in a JSON rule.
```
JohnathonMohr, line 11:
``` suggestion
    /// Represents an allOf expression in a JSON rule.
```

src/Analyzer.JsonRuleEngine/Expressions/AllOfExpression.cs
```diff
@@ -0,0 +1,80 @@
...
+    /// </summary>
+    internal class AllOfExpression : Expression
+    {
+        /// <summary>
+        /// Gets or sets the expressions to be evaluated.
```
JohnathonMohr, line 16: Remove "or sets"

src/Analyzer.JsonRuleEngine/Converters/ExpressionConverter.cs
```diff
@@ -45,17 +57,53 @@ public override ExpressionDefinition ReadJson(JsonReader reader, Type objectType
...
+                // See if a property representing a structured expression is present.  If so, parse that structured expression.
+                var structuredExpression = objectPropertyNames.FirstOrDefault(property => structuredExpressions.Contains(property));
+                if (!string.IsNullOrEmpty(structuredExpression))
+                {
+                    // Iterate over all expressions in structured expression and validate it
+                    foreach (var expression in jsonObject.InsensitiveToken(structuredExpression))
+                    {
+                        ValidateExpressions((JObject)expression, structuredExpressions, expressionJsonPropertyNames);
+                    }
+                }
```
JohnathonMohr, line 101: You don't need to do this.  ValidateExpressions will end up being called anyway as each Expression is parsed while creating the AllOf instance.

src/Analyzer.JsonRuleEngine/Converters/ExpressionConverter.cs
```diff
@@ -45,17 +57,53 @@ public override ExpressionDefinition ReadJson(JsonReader reader, Type objectType
...
+        }
+
+        private void ValidateExpressions(JObject jsonObject, HashSet<string> structuredExpressions, HashSet<string> expressionJsonPropertyNames)
+        {
+            var objectPropertyNames = jsonObject.Properties().Select(property => property.Name).ToList();
+
+            var expressionPropertyCount = objectPropertyNames.Count(property => expressionJsonPropertyNames.Contains(property));
+            if (expressionPropertyCount == 1)
+            {
+                // See if a property representing a structured expression is present.  If so, parse that structured expression.
```
JohnathonMohr, line 92: This comment should be moved to above line 69.

JohnathonMohr (changes requested):

src/Analyzer.JsonRuleEngine/Expressions/AllOfExpression.cs
```diff
@@ -0,0 +1,80 @@
...
+        /// </summary>
+        public Expression[] AllOf { get; private set; }
+
+        /// <summary>
+        /// Gets the type of resource to evaluate.
+        /// </summary>
+        public string ResourceType { get; private set; }
+
```
JohnathonMohr, line 24: `public string Path {...}` is missing?

src/Analyzer.Core/Rules/BuiltInRules.json
```diff
@@ -20,5 +20,35 @@
       "path": "properties.type",
       "notEquals": "CustomRole"
     }
+  },
+  {
+    "name": "AutomationAccounts_Variables_Encrypted",
+    "description": "Automation account variables should be encrypted",
+    "recommendation": "Enable encryption of Automation account variable assets when storing sensitive data",
+    "helpUri": "https://github.com/Azure/template-analyzer/docs/built-in-armory-rules.md/#automation-account-variables-should-be-encrypted",
```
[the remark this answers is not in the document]
yane3628: What do you mean?

src/Analyzer.JsonRuleEngine.UnitTests/AllOfExpressionDefinitionTests.cs
```diff
@@ -0,0 +1,93 @@
...
+            {
+                Path = "json.path",
+                AllOf = GenerateAllOfExpressionDefinition(numberOfExpressionDefinitions, mockLeafExpressionDefinition.Object).ToArray()
+            };
+
+            // Act
+            var allOfExpression = allOfExpressionDefinition.ToExpression() as AllOfExpression;
+
+            // Assert
+            mockLeafExpressionDefinition.Verify(s => s.ToExpression(), Times.Exactly(numberOfExpressionDefinitions));
```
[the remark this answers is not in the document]
yane3628: I use the same mock multiple times. Are you recommending making a unique mock for each leafExpression in allOf?

src/Analyzer.JsonRuleEngine.UnitTests/AllOfExpressionTests.cs
```diff
@@ -0,0 +1,111 @@
...
+namespace Microsoft.Azure.Templates.Analyzer.JsonRuleEngine.UnitTests
+{
+    [TestClass]
+    public class AllOfExpressionTests
+    {
+        [DataTestMethod]
+        [DataRow(true, true, DisplayName = "AllOf evaluates to true")]
+        [DataRow(true, false, DisplayName = "AllOf evaluates to false")]
+        [DataRow(false, false, DisplayName = "AllOf evaluates to false")]
+        [DataRow(false, false, "ResourceProvider/resource", DisplayName = "AllOf - scoped to resourceType - evaluates to true")]
```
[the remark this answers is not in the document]
yane3628: Is there anything else you recommend besides fixing the last one which should be "...evaluates to false" instead of true?

src/Analyzer.Core/Rules/BuiltInRules.json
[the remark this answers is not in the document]
JohnathonMohr: They currently reference built-in-armory-rules.md.  The file name needs updating.

src/Analyzer.JsonRuleEngine.UnitTests/AllOfExpressionDefinitionTests.cs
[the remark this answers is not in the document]
JohnathonMohr: Yes

src/Analyzer.JsonRuleEngine.UnitTests/AllOfExpressionTests.cs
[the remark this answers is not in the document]
JohnathonMohr: The middle two have the same name.  The name should demonstrate differences between each test.

src/Analyzer.Core/Rules/BuiltInRules.json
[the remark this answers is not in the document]
JohnathonMohr: (the links don't work as they are)

src/Analyzer.Core/Rules/BuiltInRules.json
```diff
@@ -20,5 +20,35 @@
...
+    "helpUri": "https://github.com/Azure/template-analyzer/docs/built-in-armory-rules.md/#role-based-access-control-rbac-should-be-used-on-kubernetes-services",
+    "evaluation": {
+      "resourceType": "Microsoft.ContainerService/managedClusters",
+      "allOf": [
+        {
+          "path": "properties.enableRBAC",
+          "hasValue": true
+        },
+        {
+          "path": "properties.enableRBAC",
```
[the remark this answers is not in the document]
yane3628: I will leave this check for now and resolve this issue in the anyOf work.

src/Analyzer.JsonRuleEngine.UnitTests/AllOfExpressionDefinitionTests.cs
```diff
@@ -0,0 +1,93 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+
+namespace Microsoft.Azure.Templates.Analyzer.JsonRuleEngine.UnitTests
+{
+    [TestClass]
+    public class AllOfExpressionDefinitionTests
+    {
+        [DataTestMethod]
+        [DataRow(0, DisplayName = "0 expressions defined in AllOf")]
```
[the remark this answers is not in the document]
yane3628: Good point. I can add this restriction to the expression converter.

src/Analyzer.JsonRuleEngine.UnitTests/JsonRuleEvaluationTests.cs
```diff
@@ -0,0 +1,160 @@
...
+                }
+                .HelpUri);
+        }
+
+        [DataTestMethod]
+        [DataRow(3, 3, 0, DisplayName = "3/3 results are true, 0/3 results are false")]
+        [DataRow(3, 2, 1, DisplayName = "2/3 results are true, 1/3 results are false")]
+        [DataRow(3, 0, 3, DisplayName = "0/3 results are true, 3/3 results are false")]
+        [DataRow(0, 0, 0, DisplayName = "No results")]
+        public void GetResultsEvaluatedTrueFalse_ThreeResults_ExpectedNumberOfResultsReturned(int totalResults, int expectedTrue, int expectedFalse)
```
JohnathonMohr, line 71: Remove parameter `expectedFalse` and just do `totalResults - expectedTrue` in the test.

JohnathonMohr: I would also suggest simplifying the display names - 3/3 true, 0/3 false is a little redundant.

src/Analyzer.JsonRuleEngine.UnitTests/JsonRuleEvaluationTests.cs
```diff
@@ -0,0 +1,160 @@
...
+            Assert.IsNull(resultsEvaluatedTrue);
+            Assert.IsNull(resultsEvaluatedFalse);
+        }
+
+        [DataTestMethod]
+        [DataRow(3, 3, 0, DisplayName = "3/3 evaluations are true, 0/3 evaluations are false")]
+        [DataRow(3, 2, 1, DisplayName = "2/3 evaluations are true, 1/3 evaluations are false")]
+        [DataRow(3, 0, 3, DisplayName = "0/3 evaluations are true, 3/3 evaluations are false")]
+        [DataRow(0, 0, 0, DisplayName = "No evaluations")]
+        public void GetEvaluationsEvaluatedTrueFalse_ThreeResults_ExpectedNumberOfEvaluationsReturned(int totalEvaluations, int expectedTrue, int expectedFalse)
```
JohnathonMohr, line 118: ThreeEvaluations?

src/Analyzer.Types/IEvaluation.cs
```diff
@@ -0,0 +1,63 @@
...
+        /// </summary>
+        public IEnumerable<IResult> Results { get; }
+
+        /// <summary>
+        /// Gets wether this evaluation has corresponding results.
+        /// </summary>
+        /// <returns>Whether this evaluation has corresponding results</returns>
+        public bool HasResults()
+        {
+            return Results != null && Results.Count() > 0;
```
JohnathonMohr, line 60: I don't know if it matters right now, but this will have the same IEnumerable iteration issue we discussed earlier - `.Count()` will iterate over it, so the next time it's referenced, it could reevaluate everything again.

JohnathonMohr (changes requested):

src/Analyzer.JsonRuleEngine.UnitTests/AllOfExpressionTests.cs
```diff
@@ -0,0 +1,121 @@
...
+
+            var allOfExpression = new AllOfExpression(expressionArray, resourceType: resourceType, path: path);
+
+            // Act
+            var allOfEvaluation = allOfExpression.Evaluate(mockJsonPathResolver.Object);
+
+            // Assert
+            bool expectedAllOfEvaluation = evaluation1 && evaluation2;
+            Assert.AreEqual(expectedAllOfEvaluation, allOfEvaluation.Passed);
+            Assert.IsFalse((allOfEvaluation as IEvaluation).HasResults());
```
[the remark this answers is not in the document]
yane3628: Because HasResults() is defined in IEvaluation and not JsonRuleEvaluation, there is a compilation error if the cast isn't present.

src/Analyzer.JsonRuleEngine/JsonRuleEngine.cs
```diff
@@ -33,33 +34,39 @@ public JsonRuleEngine(IJsonLineNumberResolver lineNumberResolver)
...
                 var ruleExpression = rule.Evaluation.ToExpression();
-                var ruleResults = ruleExpression.Evaluate(
+                var evaluation = ruleExpression.Evaluate(
                     new JsonPathResolver(
                         templateContext.ExpandedTemplate,
                         templateContext.ExpandedTemplate.Path));
 
-                foreach (var result in ruleResults)
+                // If there are no matching cases of the rule, do not create an evaluation
+                if (!(evaluation as IEvaluation).HasResults())
```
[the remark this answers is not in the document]
yane3628: This is here because it was creating empty evaluations for rules where even the resource type was not present. Without this, there will always be one evaluation result for each rule even if the rule doesn't apply to the template.

src/Analyzer.Types/IEvaluation.cs
yane3628: I tried making it a property instead and cache the bool, but it won't work since the class is an interface.

src/Analyzer.JsonRuleEngine.UnitTests/AllOfExpressionTests.cs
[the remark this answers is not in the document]
JohnathonMohr: Interesting, this might be a new feature of C# 8 (having an implementation in an Interface).

src/Analyzer.JsonRuleEngine/JsonRuleEngine.cs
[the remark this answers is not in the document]
JohnathonMohr: That makes sense.  But I think there's still a problem in the allOf scenario - allOf will never have results.

src/Analyzer.Types/IEvaluation.cs
JohnathonMohr: Maybe this should be moved into JsonRuleEvaluation.  We can take the decision later.

src/Analyzer.JsonRuleEngine/JsonRuleEngine.cs
```diff
@@ -40,26 +41,56 @@ public IEnumerable<IEvaluation> EvaluateRules(TemplateContext templateContext, s
...
                     new JsonPathResolver(
                         templateContext.ExpandedTemplate,
                         templateContext.ExpandedTemplate.Path));
 
+                 evaluation.RuleDefinition = rule;
+                 evaluation.FileIdentifier = templateContext.TemplateIdentifier;
+
                 // If there are no matching cases of the rule, do not create an evaluation
-                if (!(evaluation as IEvaluation).HasResults())
+                if (rule.ExpressionDefinition is LeafExpressionDefinition)
```
JohnathonMohr, line 54: Instead of looking at what kind of expression was evaluated, can you make HasResults recursive?  i.e. HasResults could look at the Results collection, and if that's empty, will see if any of its Evaluations have results.  That way the rule engine doesn't have to understand the details of how rules are evaluated, and I think will be less code overall.

src/Analyzer.JsonRuleEngine.UnitTests/JsonRuleEngineTests.cs
```diff
@@ -77,16 +78,143 @@ public void Run_ValidInputs_ReeturnsExpectedResults(string[] evaluations, int nu
...
+                    It.Is<JToken>(templateContext.ExpandedTemplate, EqualityComparer<object>.Default),
+                    It.Is<JToken>(templateContext.OriginalTemplate, EqualityComparer<object>.Default)))
+                .Returns(expectedLineNumber);
+
+            var ruleEngine = new JsonRuleEngine(mockLineResolver.Object);
+            var evaluationResults = ruleEngine.EvaluateRules(templateContext, rules).ToList();
+
+            Assert.AreEqual(1, evaluationResults.Count());
+            Assert.AreEqual(1, evaluationResults.Count(evaluation => evaluation.Passed));
+            for (int i = 0; i < 1; i++)
```
JohnathonMohr, line 191: for-loop probably not needed.

JohnathonMohr (approved):
I think these changes will work well enough for us to keep going.  I have 1 suggestion for how Evaluations/Results can be improved, but it can be done later if you choose.

PR #61, opened 2021-03-18
Refactoring Evaluation Logic

JohnathonMohr:
Refactoring how Expressions are evaluated and when a line number is generated for results.  This will provide a base for adding the Where clause into the rule language.

- Pushing `JsonLineNumberResolver` down into `LeafExpression`.
  - `LeafExpression` now assigns a line number when creating the result directly
  - `JsonRuleEngine` no longer needs to iterate through results to assign a line number
- Extracted common logic of expanding JSON scopes and combining results in various Expressions into `Expression` base class
  - Child classes call base class for expanding scope
  - Now they only need to worry about how to evaluate themselves on a given scope
- Some cleanup in JsonRuleEvaluation
  - `Evaluations` and `Results` always have a valid, non-null Enumerable, with 0 elements if not originally set
  - `HasResults` is recursively looks through sub-Evaluations to determine if any results exist at all
  - `JsonRuleEngine` can call `HasResults` just once to determine if an Evaluation has anything that was evaluated in it
- Test updates for changes in method/constructor signatures, and slight behavior change in `JsonRuleEvaluation`


src/Analyzer.JsonRuleEngine/Expressions/Expression.cs
```diff
@@ -10,11 +14,114 @@ namespace Microsoft.Azure.Templates.Analyzer.RuleEngines.JsonEngine.Expressions
...
+            Func<IJsonPathResolver, JsonRuleEvaluation> getEvaluation = null,
+            Func<IJsonPathResolver, JsonRuleResult> getResult = null)
+        {
+            if (jsonScope == null)
+            {
+                throw new ArgumentNullException(nameof(jsonScope));
+            }
+            if (getEvaluation == null && getResult == null)
+            {
+                throw new ArgumentNullException($"{nameof(getEvaluation)} and {nameof(getResult)}");
```
yane3628, line 69: If you don't refactor this function, add tests for these ArgumentNullExceptions

src/Analyzer.JsonRuleEngine/Expressions/LeafExpression.cs
```diff
@@ -44,49 +38,29 @@ public LeafExpression(string resourceType, string path, LeafExpressionOperator @
...
-                    });
-                }
-            }
+                    };
+
+                    try
+                    {
+                        result.LineNumber = this.jsonLineNumberResolver.ResolveLineNumber(result.JsonPath);
+                    }
+                    catch (Exception) { }
```
yane3628, line 60: Can a test be added for this catch statement?

src/Analyzer.JsonRuleEngine.FunctionalTests/RuleParsingTests.cs
```diff
@@ -61,7 +63,7 @@ private LeafExpressionOperator ParseJsonValidateAndReturnOperator(string operato
                 operatorProperty,
                 jsonValue));
 
-            var leafExpression = leafDefinition.ToExpression() as LeafExpression;
+            var leafExpression = leafDefinition.ToExpression(new Mock<ILineNumberResolver>().Object) as LeafExpression;
```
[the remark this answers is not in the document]
JohnathonMohr: These tests are only verifying rule parsing, so line number resolution is not needed and shouldn't be evaluated.

src/Analyzer.JsonRuleEngine/JsonRuleEngine.cs
```diff
@@ -17,16 +17,23 @@ namespace Microsoft.Azure.Templates.Analyzer.RuleEngines.JsonEngine
     /// </summary>
     public class JsonRuleEngine : IRuleEngine
     {
-        private readonly IJsonLineNumberResolver lineNumberResolver;
+        /// <summary>
+        /// Delegate for building an <c>IJsonLineNumberResolver</c>
```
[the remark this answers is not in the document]
JohnathonMohr: I started using <cref /> and then forgot to go back and update 😣

src/Analyzer.Utilities/IJsonLineNumberResolver.cs
```diff
@@ -1,28 +1,18 @@
...
-using Newtonsoft.Json.Linq;
-
 namespace Microsoft.Azure.Templates.Analyzer.Utilities
 {
     /// <summary>
-    /// An interface used for resolving line numbers from an expanded template to the original template.
+    /// An interface used for resolving line numbers in a document.
     /// </summary>
-    public interface IJsonLineNumberResolver
+    public interface ILineNumberResolver
```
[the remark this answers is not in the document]
JohnathonMohr: Good catch!

src/Analyzer.JsonRuleEngine/Expressions/Expression.cs
JohnathonMohr: Refactored.

src/Analyzer.JsonRuleEngine/Expressions/LeafExpression.cs
JohnathonMohr: I decided this was not necessary.  If this does ever throw an exception, it will be because of a bug in rule evaluation setup, not something wrong in `LeafExpression`.

yane3628 (approved):

PR #82, opened 2021-04-10
Support for wildcards in path specification

### LINKED ISSUE #14: Add support for wildcard in a JArray
The `path` field can support a wildcard to query all elements in a JArray 
`somePath.array[*]`

### LINKED ISSUE #85: Support wildcard property names in path property of rule evaluation
In a rule `Evaluation`, the `path` property should handle a wildcard as a property name in the JSON path.  This should result in all JSON paths that satisfy the specified path value being returned for evaluation.

### Template JSON to evaluate
``` js
{
  "prop1": {
    "child1": {
      "value1": 1,
      "value2": 2
    },
    "child2": {
      "value1": 1,
      "value3": 3
    },
    "child3": {
      "value1": 1,
      "value4": 4
    }
  }
}
```
### Rule Evaluation
``` js
{
  "path": "prop1.*.value1"
  // ...
}
```

### Paths evaluated
`prop1.child1.value1`
`prop1.child2.value1`
`prop1.child3.value1`

JohnathonMohr:
These changes implement wildcard resolution for the `path` specified in `Expression`s.  Wildcards can be specified for both a property name and an array index.  (Closes #14.  Closes #85)
- Implemented `JTokenExtensions.InsensitiveTokens()` which supports wildcards in path
  - Returns an Enumerable of JTokens (all JTokens which match the path specification)
  - As the function looks for the next property in the template to match against the next property in the path, it adds a match to a queue.
    - If a wildcard is encountered in the path, it adds all children and/or all array elements to queue for processing.
  - Each item in queue has its own context for searching for the next property in the path.
- All `JTokenExtensions.InsensitiveToken()` implementations now rely on the new `Tokens()` method
- `JsonPathResolver` calls `InsensitiveTokens` for all path resolution
- Added unit tests for wildcard selection
- Added unit tests for wildcard resolving in `JsonPathResolver`
- Added functional tests for validating the scopes evaluated by an `Expression`.
- Updated documentation


src/Analyzer.Utilities/JsonLineNumberResolver.cs
```diff
@@ -55,7 +55,15 @@ public int ResolveLineNumber(string pathInExpandedTemplate)
...
-                // See if the path starts with indexing into the template resources[] array
+                return 0;
+            }
+
+            // If the JToken returned from looking up the expanded template path is
+            // pointing to the resources array of the original template, then it's likely
+            // the expanded template path used an index too large for the original template.
+            if (tokenFromOriginalTemplate.Equals(originalTemplateRoot.InsensitiveToken("resources")))
+            {
+                // Verify the path starts with indexing into the template resources[] array
```
JohnathonMohr, line 66: This small code change was needed because of the algorithm change when accounting for wildcards.

Previously (for example), if the first property in the expanded path was *resources[5]* and the original template only had resources 0-3, `originalTemplateRoot.InsensitiveToken()` would fail to find exactly *resources[5]*, and so would return the root node.

Now, `InsensitiveToken()` will find *resources*, but not *element 5*, and will return the node representing the whole *resources* array.

yane3628 (changes requested):

docs/authoring-json-rules.md
```diff
@@ -383,4 +383,9 @@ In the simple example above, the `evaluate` operator would be skipped, because t
...
\ No newline at end of file
+**NOTE:** In both examples above, `"path": "properties.osProfile.computerName"` is specified *inside* the `evaluate` operator.  This is important because of how [scopes](#scopes) are determined.  If it was instead specified outside the operator (as a sibling to `where`), it would narrow the outer scope to that path.  That path would then be passed into `where`, resulting in `"path": "apiVersion"` and `"path": "name"` (inside `where` in the examples) being appended to *properties.osProfile.computerName* in the outer scope, which is not the intent.
+
+## Wildcard Behavior
+The `path` in an `Evaluation` can specify the '\*' character as a wildcard.  '\*' can be used to match any full property name or as the index into an array (selecting all elements of the array).  When a wildcard is used, zero or more paths in the template will be found that match `path`.  If zero are found, the operator in the `Evaluation` is skipped, as there is nothing to evaluate.  If two or more are found, the operator evaluates each path individually and the results are logically 'and'ed together.
```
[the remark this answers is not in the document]
JohnathonMohr: Italicizing paths is a good idea.  I'm not sure about * (regular) vs *\** (in italics).  I think single quotes around the character are still helpful.

src/Analyzer.Utilities.UnitTests/JTokenExtensionsTests.cs
```diff
@@ -10,187 +13,472 @@ namespace Microsoft.Azure.Templates.Analyzer.Utilities.UnitTests
...
-            bool bothValuesAreEqual = JToken.DeepEquals(parsedJson, valueAtPath);
+        /// <summary>
+        /// Test data for tests that verify paths with wildcards, where potentially multiple tokens could be returned.
+        /// Index 1: The path to test.
+        /// Index 2: A list of object arrays, each array being the literal child elements to an expected token that will be matched from the path being tested.
+        ///          A list element containing an empty array, followed by more elements of non-empty arrays,
+        ///          denotes literal child elements to an expected token returned with LastValid behavior.
+        /// Index 3: The display name of the test.
+        /// </summary>
+        public static IReadOnlyList<object[]> WildcardTestScenarios { get; } = new List<object[]>
```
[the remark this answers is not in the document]
JohnathonMohr: That's not supported yet.  Just '\*' will return the values of all top-level properties, but not recursively select all their children.  To select every child path, we still need to define that behavior.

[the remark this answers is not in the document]
JohnathonMohr: I could add a test for just '\*' though.

VeraBE (approved):

src/Analyzer.JsonRuleEngine.FunctionalTests/ScopeSelectionTests.cs
```diff
@@ -0,0 +1,254 @@
...
+            DisplayName = "Resource type matches multiple, path not resolved")]
+        [DataRow("Microsoft.Network/networkInterfaces", "properties.networkSecurityGroup",
+            "resources[1].properties.networkSecurityGroup", null,
+            DisplayName = "Resource type matches multiple, path resolves in 1")]
+        [DataRow("Microsoft.Network/networkInterfaces", "properties.ipConfigurations[0]",
+            "resources[1].properties.ipConfigurations[0]", "resources[2].properties.ipConfigurations[0]",
+            DisplayName = "Resource type matches multiple, path resolves in all")]
+        [DataRow("Microsoft.Compute/virtualMachines", "properties.hardwareProfile.*",
+            // No scopes expected evaluated
+            DisplayName = "Resource type matches multiple, wildcard path does not resolve")]
```
anaismiller, line 224: And is this resource type correct?

src/Analyzer.JsonRuleEngine.FunctionalTests/ScopeSelectionTests.cs
```diff
@@ -0,0 +1,254 @@
...
+            DisplayName = "Resource type matches none, path not resolved")]
+        [DataRow("Microsoft.Network/virtualNetworks", "properties.addressSpace",
+            null,
+            DisplayName = "Resource type matches 1, path not resolved")]
+        [DataRow("Microsoft.Network/virtualNetworks", "location",
+            "resources[0].location",
+            DisplayName = "Resource type matches 1, path resolved")]
+        [DataRow("Microsoft.Network/virtualNetworks", "dependsOn[*]",
+            // No scopes expected evaluated
+            DisplayName = "Resource type matches 1, wildcard path does not resolve")]
```
anaismiller, line 206: Maybe I don't understand the logic, but shouldn't this be `"Resource type matches none"` as you gave no scope (like line  197)?

src/Analyzer.JsonRuleEngine.FunctionalTests/ScopeSelectionTests.cs
```diff
@@ -0,0 +1,254 @@
...
+            DisplayName = "Resource type matches 1, no path")]
+        [DataRow("Microsoft.Compute/virtualMachines", null,
+            "resources[3]", "resources[4]",
+            DisplayName = "Resource type matches multiple, no path")]
+        [DataRow("Microsoft.Storage/storageAccounts", "name",
+            // No scopes expected evaluated
+            DisplayName = "Resource type matches none, path not resolved")]
+        [DataRow("Microsoft.Network/virtualNetworks", "properties.addressSpace",
+            null,
+            DisplayName = "Resource type matches 1, path not resolved")]
```
anaismiller, line 200: is null not the same thing as not passing in a parameter/ why did you decide not (also line 214)?

src/Analyzer.Utilities/JTokenExtensions.cs
```diff
@@ -13,107 +15,320 @@ namespace Microsoft.Azure.Templates.Analyzer.Utilities
...
+
+                                            if (intIndex < childArray.Count)
+                                            {
+                                                // If index is in-bounds, select that index.
+                                                EnqueueChildJToken(jtokensToSearchFrom, childArray[intIndex], tokenContext); 
+                                            }
+                                            else
+                                            {
+                                                // Otherwise, set the context to the array itself
+                                                // and set found to false so searching stops.
```
anaismiller, line 224: I don't see how by setting found to false, it stops the searching (ie. where in this foreach loop does it make the search stop). Maybe I am not looking well. Also, could be better to put line 197 `foundChild = true;` in the if condition instead if you are just declaring the bool in line 188, then later putting true, then putting false. But again, I may be missing something.

src/Analyzer.Utilities/JTokenExtensions.cs
```diff
@@ -13,107 +15,320 @@ namespace Microsoft.Azure.Templates.Analyzer.Utilities
...
-                                var index = int.Parse(arrayMatch.Groups["index"].Value);
-                                token = childProperty.Value[index];
-                                foundChild = true;
+                                // Path is indexing into this array property.
+                                // Attempt to find the requested index.
+                                if (childProperty.Value is JArray childArray)
+                                {
+                                    var index = arrayMatch.Groups["index"].Value;
+                                    if (index == "*")
+                                    {
```
anaismiller, line 206: Could be better to put line 197 foundChild = true; in the if condition instead here as you are just declaring the bool in line 188, then later putting true, then putting false if the condition breaks.

src/Analyzer.Utilities/JTokenExtensions.cs
```diff
@@ -13,107 +15,320 @@ namespace Microsoft.Azure.Templates.Analyzer.Utilities
...
+                    EnqueueChildJToken(tokenQueue, childProperty.Value, tokenContext);
                 }
             }
+        }
+
+        /// <summary>
+        /// Helper for InsensitiveTokens().
+        /// Adds all elements of a JArray to the queue to be processed later.
+        /// </summary>
+        private static void AddAllArrayElementsToQueue(Queue<InsensitiveTokenContext> tokenQueue, JArray jArray, InsensitiveTokenContext parentContext)
```
anaismiller, line 307: Do you want to call this `AddAllJArrayElementsToQueue` instead?

anaismiller (approved):
approve with questions/suggestions.

src/Analyzer.JsonRuleEngine.FunctionalTests/ScopeSelectionTests.cs
JohnathonMohr: Yes - what's the reason you're unsure?  Is it the path not matching anything in the resources?  It's not supposed to.

src/Analyzer.JsonRuleEngine.FunctionalTests/ScopeSelectionTests.cs
JohnathonMohr: The resource type specified does exist in the template (there is one resource of that type), but the path specified with the wildcard doesn't match any path, so no path is resolved.

src/Analyzer.JsonRuleEngine.FunctionalTests/ScopeSelectionTests.cs
JohnathonMohr: Maybe I should update the comments.  If a static path isn't found, null is evaluated as the path value (since it isn't found).  If a wildcard path isn't found, it returns empty, and nothing is evaluated.

src/Analyzer.Utilities/JTokenExtensions.cs
JohnathonMohr: It looks like no matter how I do it (setting `true` only when absolutely sure it's true, or setting `false` if a condition breaks), I have to explicitly set `foundChild` to something in three separate places (either three `true`s or three `false`s).  So I guess it's just preference.  Unless you're feeling strongly one way, I think I'll leave it as-is.

src/Analyzer.Utilities/JTokenExtensions.cs
JohnathonMohr: I can do that.

src/Analyzer.Utilities.UnitTests/JTokenExtensionsTests.cs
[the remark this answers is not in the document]
yane3628: We should add the recursive bit as a issue. Probably something like *.** or something.

yane3628 (approved):

PR #84, opened 2021-04-12
Add support for regex operator

### LINKED ISSUE #18: Add support for regex comparisons
Runs the regular expression in the specified value against the value at the path
``` javascript 
{
    "regex": "admin" // Evaluates to `false` because "admin" is not contained in the value of the path "properties.osProfile.adminUsername".
}
```

### LINKED ISSUE #76: [Built-in Rule] API App should only be accessible over HTTPS
Azure Policy link: https://github.com/Azure/azure-policy/blob/master/built-in-policies/policyDefinitions/App%20Service/AppServiceApiApp_AuditHTTP_Audit.json

``` js
{
  "name": "AppServiceApiApp_HTTPS",
  "description": "API App should only be accessible over HTTPS",
  "recommendation": "Use HTTPS to ensure server/service authentication and protect data in transit from network layer eavesdropping attacks.",
  "helpUri": "https://github.com/Azure/template-analyzer/docs/built-in-bpa-rules.md/#api-app-should-only-be-accessible-over-https",
  "evaluation": {
    "resourceType": "Microsoft.Web/sites",
    "where": {
      "path": "kind",
      "regex": "api$"
    },
    "evaluate": {
      "path": "properties.httpsOnly",
      "equals": true
    }
  }
}
```

yane3628:
Add support of regex operator (Closes #18)
Add tests for regex
Update documentation
Add built in rule: API App should only be accessible over HTTPS (Closes #76)

### DIFF
--- a/docs/authoring-json-rules.md
+++ b/docs/authoring-json-rules.md
@@ -198,7 +198,7 @@ Example:
 }
 ```
 
-#### **Regex [[*]](#note)**
+#### **Regex**
 *Type: string*
 
 Runs the regular expression in the specified value against the value of the `path` in the template.  All regular expressions are case-insensitive.  Evaluates to `true` if the regular expression is a match; `false` otherwise.  If the value in the template is not a string, this evaluates to `false`.
--- a/src/Analyzer.Core/Rules/BuiltInRules.json
+++ b/src/Analyzer.Core/Rules/BuiltInRules.json
@@ -22,44 +22,6 @@
     }
   },
   {
-    "name": "Compute_VirtualMachine_ShouldNotBePreview",
-    "description": "Ensures that all virtual machines are not using preview images",
-    "recommendation": "Use a non-preview image for virtual machines",
-    "helpUri": "https://github.com/Azure/template-analyzer/docs/built-in-bpa-rules.md/#ensures-that-all-virtual-machines-are-not-using-preview-images",
-    "evaluation": {
-      "resourceType": "Microsoft.Compute/virtualMachines",
-      "anyOf": [
-        {
-          "path": "properties.storageProfile.imageReference",
-          "regex": "^((?!-preview$).)*$"
-        },
-        {
-          "path": "properties.storageProfile.imageReference.version",
-          "regex": "^((?!-preview$).)*$"
-        }
-      ]
-    }
-  },
-  {
-    "name": "Compute_VirtualMachineScaleSets_ShouldNotBePreview",
-    "description": "Ensure all virtual machine scale set instances are not using preview images",
-    "recommendation": "Use a non-preview image for virtual machine scale set instances",
-    "helpUri": "https://github.com/Azure/template-analyzer/docs/built-in-bpa-rules.md/#ensure-all-virtual-machine-scale-set-instances-are-not-using-preview-images",
-    "evaluation": {
-      "resourceType": "Microsoft.Compute/virtualMachineScaleSets",
-      "anyOf": [
-        {
-          "path": "properties.virtualMachineProfile.storageProfile.imageReference",
-          "regex": "^((?!-preview$).)*$"
-        },
-        {
-          "path": "properties.virtualMachineProfile.storageProfile.imageReference.version",
-          "regex": "^((?!-preview$).)*$"
-        }
-      ]
-    }
-  },
-  {
     "name": "Kubernetes_Must_Restrict_Network_Access",
     "description": "Authorized IP ranges should be defined on Kubernetes Services",
     "recommendation": "Restrict access to the Kubernetes Service Management API by granting API access only to IP addresses in specific ranges",
--- a/src/Analyzer.Core/Rules/BuiltInRules.json
+++ b/src/Analyzer.Core/Rules/BuiltInRules.json
@@ -22,6 +22,44 @@
     }
   },
   {
+    "name": "Compute_VirtualMachine_ShouldNotBePreview",
+    "description": "Ensures that all virtual machines are not using preview images",
+    "recommendation": "Use a non-preview image for virtual machines",
+    "helpUri": "https://github.com/Azure/template-analyzer/docs/built-in-bpa-rules.md/#ensures-that-all-virtual-machines-are-not-using-preview-images",
+    "evaluation": {
+      "resourceType": "Microsoft.Compute/virtualMachines",
+      "anyOf": [
+        {
+          "path": "properties.storageProfile.imageReference",
+          "regex": "^((?!-preview$).)*$"
+        },
+        {
+          "path": "properties.storageProfile.imageReference.version",
+          "regex": "^((?!-preview$).)*$"
+        }
+      ]
+    }
+  },
+  {
+    "name": "Compute_VirtualMachineScaleSets_ShouldNotBePreview",
+    "description": "Ensure all virtual machine scale set instances are not using preview images",
+    "recommendation": "Use a non-preview image for virtual machine scale set instances",
+    "helpUri": "https://github.com/Azure/template-analyzer/docs/built-in-bpa-rules.md/#ensure-all-virtual-machine-scale-set-instances-are-not-using-preview-images",
+    "evaluation": {
+      "resourceType": "Microsoft.Compute/virtualMachineScaleSets",
+      "anyOf": [
+        {
+          "path": "properties.virtualMachineProfile.storageProfile.imageReference",
+          "regex": "^((?!-preview$).)*$"
+        },
+        {
+          "path": "properties.virtualMachineProfile.storageProfile.imageReference.version",
+          "regex": "^((?!-preview$).)*$"
+        }
+      ]
+    }
+  },
+  {
     "name": "Kubernetes_Must_Restrict_Network_Access",
     "description": "Authorized IP ranges should be defined on Kubernetes Services",
     "recommendation": "Restrict access to the Kubernetes Service Management API by granting API access only to IP addresses in specific ranges",
--- a/src/Analyzer.Core/Rules/BuiltInRules.json
+++ b/src/Analyzer.Core/Rules/BuiltInRules.json
@@ -1,5 +1,24 @@
 [
   {
+    "name": "AppServiceApiApp_HTTPS",
+    "description": "API App should only be accessible over HTTPS",
+    "recommendation": "Use HTTPS to ensure server/service authentication and protect data in transit from network layer eavesdropping attacks.",
+    "helpUri": "https://github.com/Azure/template-analyzer/docs/built-in-bpa-rules.md/#api-app-should-only-be-accessible-over-https",
+    "evaluation": {
+      "resourceType": "Microsoft.Web/sites",
+      "where": {
+        "path": "kind",
+        "regex": "api$"
+      },
+      "allOf": [
+        {
+          "path": "properties.httpsOnly",
+          "equals": true
+        }
+      ]
+    }
+  },
+  {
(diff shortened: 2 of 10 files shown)


src/Analyzer.JsonRuleEngine/Operators/RegexOperator.cs
```diff
@@ -0,0 +1,47 @@
...
+        /// Gets the regex pattern this operator will compare against.
+        /// </summary>
+        public string RegexPattern { get; private set; }
+
+        /// <summary>
+        /// Creates a RegexOperator.
+        /// </summary>
+        /// <param name="regexPattern">The regex pattern specified in the JSON rule.</param>
+        /// <param name="isNegative">Whether the result of EvaluateExpression() should be negated or not.</param>
+        public RegexOperator(string regexPattern, bool isNegative)
```
JohnathonMohr, line 24: I think for regex it doesn't make sense to have an `isNegative` parameter.  This can be removed, and `this.IsNegative` can just always be `false`.

src/Analyzer.JsonRuleEngine/Operators/RegexOperator.cs
```diff
@@ -0,0 +1,47 @@
...
+        /// <param name="tokenToEvaluate">The JToken to evaluate.</param>
+        /// <returns>A value indicating whether or not the evaluation passed.</returns>
+        public override bool EvaluateExpression(JToken tokenToEvaluate)
+        {
+            if (tokenToEvaluate.Type != JTokenType.String)
+            {
+                return false;
+            }
+
+            Regex regex = new Regex(this.RegexPattern, RegexOptions.IgnoreCase);
```
JohnathonMohr, line 43: Move this to the constructor, and create the Regex with `RegexOptions.Compiled` as well.

src/Analyzer.JsonRuleEngine.UnitTests/RegexOperatorTests.cs
```diff
@@ -0,0 +1,60 @@
...
+            // {"Regex": jTokenValue} is false
+            var regexOperator = new RegexOperator(regex, isNegative: false);
+            Assert.IsFalse(regexOperator.EvaluateExpression(stringToMatch));
+        }
+
+        [DataTestMethod]
+        [DataRow(0, DisplayName = "Property is an int")]
+        [DataRow(false, DisplayName = "Property is a boolean")]
+        [DataRow(0.1, DisplayName = "Property is a float")]
+        [DataRow(new string[] { "value" }, DisplayName = "Property is an array")]
```
JohnathonMohr, line 43: Add a row for `null`

JohnathonMohr: Two rows might be valuable here:
- simply `null`
- and `(string)null`

JohnathonMohr (changes requested):
Looks good overall.  Just a couple comments.

src/Analyzer.Core/Rules/BuiltInRules.json
```diff
@@ -21,6 +21,44 @@
...
     }
   },
+  {
+    "name": "Compute_VirtualMachine_ShouldNotBePreview",
+    "description": "Ensures that all virtual machines are not using preview images",
+    "recommendation": "Use a non-preview image for virtual machines",
+    "helpUri": "https://github.com/Azure/template-analyzer/docs/built-in-bpa-rules.md/#ensures-that-all-virtual-machines-are-not-using-preview-images",
+    "evaluation": {
+      "resourceType": "Microsoft.Compute/virtualMachines",
+      "anyOf": [
```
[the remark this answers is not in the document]
anaismiller: Why? (not familiar with looking at these, but isn't it possible that only the version or only the reference has the preview?)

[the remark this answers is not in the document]
JohnathonMohr: I believe the regex enforces that "-preview" *does not* appear in the string.  So the rule should say "both paths do not have -preview".

docs/authoring-json-rules.md
```diff
@@ -198,7 +198,7 @@ Example:
 }
 ```
 
-#### **Regex [[*]](#note)**
+#### **Regex**
 *Type: string*
 
 Runs the regular expression in the specified value against the value of the `path` in the template.  All regular expressions are case-insensitive.  Evaluates to `true` if the regular expression is a match; `false` otherwise.  If the value in the template is not a string, this evaluates to `false`.
```
anaismiller, line 204: On line 274, please change "anyOf" to "allOf" in the example.

src/Analyzer.JsonRuleEngine.UnitTests/RegexOperatorTests.cs
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        [TestMethod]
+        public void GetName_ReturnsCorrectName()
+        {
+            Assert.AreEqual("Regex", new RegexOperator("", false).Name);
+        }
+
+        // Creates JSON with 'value' as the value of a key, parses it, then selects that key.
+        private static JToken ToJToken(object value)
```
anaismiller, line 57: Don't you all create this function in multiple tests? Would it be better to create a Constants in the test files with this function (and any others that appear often)?

JohnathonMohr: It hasn't been a big concern for us.  I think because it's such a small piece of code, only for testing, and there isn't much else shared between tests.

docs/authoring-json-rules.md
JohnathonMohr: Good catch!  I can fix this in the wildcard PR.

VeraBE (approved):

src/Analyzer.Core/Rules/BuiltInRules.json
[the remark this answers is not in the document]
anaismiller: ah, negation!

anaismiller (approved):
approving with recommendations + Johnathon's suggestions

src/Analyzer.Core/Rules/BuiltInRules.json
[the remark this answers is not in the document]
yane3628: The issue is that when storageProfile.imageReference.version is defined, then imageReference is not a string. Regex returns false since it's not a string and then allOf is false. I have been trying to find an example where imageReference matches the regex, but I haven't found anything.

src/Analyzer.JsonRuleEngine/Operators/RegexOperator.cs
yane3628: I assumed we were going to use isNegative for when the not operator is implemented.

src/Analyzer.Core/Rules/BuiltInRules.json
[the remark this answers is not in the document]
JohnathonMohr: If either field can be a string sometimes, but not always, then the rule probably needs more logic.

src/Analyzer.JsonRuleEngine/Operators/RegexOperator.cs
JohnathonMohr: I think `not` has to be its own `Expression` that 'not's the expression inside it.  The language supports doing something like `"not": { "allOf": [...] }` too, which wouldn't have any way to 'negate' the `allOf` other than having `not` produce its own Evaluation.

yane3628: I guess it also makes sense since technically doing not { notEquals } would make the isNegative property more complicated.

src/Analyzer.JsonRuleEngine.UnitTests/RegexOperatorTests.cs
```diff
@@ -0,0 +1,62 @@
...
+        [DataRow("^value", "aaavalue", DisplayName = "String does not begin with \"value\"")]
+        [DataRow("value$", "valueaaa", DisplayName = "String does not end with \"value\"")]
+        [DataRow("^((?!value).)*$", "value", DisplayName = "String contains \"value\"")]
+        public void EvaluateExpression_StringDoesNotMatchRegex_EvaluationIsFalse(string regex, string stringToMatch)
+        {
+            // {"Regex": jTokenValue} is false
+            var regexOperator = new RegexOperator(regex, isNegative: false);
+            Assert.IsFalse(regexOperator.EvaluateExpression(stringToMatch));
+        }
+
```
JohnathonMohr, line 38: Add a test with the `regex` operator having an invalid regular expression (like a character group or match group not having a closing bracket, or escaping an invalid character, or something).

yane3628: What is the expected behavior here? I assume we'd want to just throw the exception so they can fix the rule right?

src/Analyzer.JsonRuleEngine/JsonRuleEngine.cs
```diff
@@ -42,11 +43,30 @@ public JsonRuleEngine(BuildIJsonLineNumberResolver jsonLineNumberResolverBuilder
...
 
-            foreach(RuleDefinition rule in rules)
+            try
             {
-                var ruleExpression = rule.ExpressionDefinition.ToExpression(BuildLineNumberResolver(templateContext));
+                rules = JsonConvert.DeserializeObject<List<RuleDefinition>>(ruleDefinitions);
+            }
+            catch (Exception e)
+            {
+                throw new JsonRuleEngineException($"Failed to parse rules {e.Message}");
```
JohnathonMohr, line 54: For these exceptions, instead of appending `e.Message`, add `e` as the `innerException`.

src/Analyzer.JsonRuleEngine/JsonRuleEngineException.cs
```diff
@@ -0,0 +1,46 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Runtime.Serialization;
+
+namespace Microsoft.Azure.Templates.Analyzer.RuleEngines.JsonEngine
+{
+    /// <summary>
+    /// Template Analyzer Exception is thrown when there was an issue in Analyzer.Core
```
JohnathonMohr, line 10: Update summary comments in file

src/Analyzer.JsonRuleEngine/Operators/RegexOperator.cs
```diff
@@ -27,7 +27,15 @@ public RegexOperator(string regexPattern)
...
             this.RegexPattern = regexPattern;
-            this.regex = new Regex(this.RegexPattern, RegexOptions.IgnoreCase | RegexOptions.Compiled);
+
+            try
+            {
+                this.regex = new Regex(this.RegexPattern, RegexOptions.IgnoreCase | RegexOptions.Compiled);
+            }
+            catch (System.ArgumentException e)
+            {
+                throw new System.ArgumentException($"Regex pattern is not valid: {e.Message}");
```
JohnathonMohr, line 37: Pass `e` as the `innerException`

JohnathonMohr (changes requested):

JohnathonMohr (approved):

src/Analyzer.Core/Rules/BuiltInRules.json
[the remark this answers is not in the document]
yane3628: Removed rule for now.

PR #98, opened 2021-04-20
Add support for 'in' operator

### LINKED ISSUE #17: Add support for the in (a list) operator
Checks whether there is an element in the array equal to the specified value
``` javascript
{
    "path": "apiVersion",
    "in": [
         "2020-06-01",
         "2019-12-01",
         "2019-07-01",
         "2019-03-01"
    ] // Evaluates to `true` because the value of "apiVersion" in the template ("2020-06-01") is equal to one of the values in the array specified for `in`.
}
```

VeraBE:
Closes #17

### DIFF
--- a/docs/authoring-json-rules.md
+++ b/docs/authoring-json-rules.md
@@ -222,7 +222,7 @@ Example:
 }
 ```
 
-#### **In [[*]](#note)**
+#### **In**
 *Type: array of basic JSON values (integer, float, string, bool, null)*
 
 Evaluates the value of the `path` in the template using the `equals` operator for each value specified in the array.  If any results in `true`, `in` will evaluate to true; `false` otherwise.  All values in the array must be of the same type.
--- a/docs/authoring-json-rules.md
+++ b/docs/authoring-json-rules.md
@@ -223,7 +223,7 @@ Example:
 ```
 
 #### **In**
-*Type: array of basic JSON values (integer, float, string, bool, null)*
+*Type: mixed-type array of basic JSON values (integer, float, string, bool, null)*
 
 Evaluates the value of the `path` in the template using the `equals` operator for each value specified in the array.  If any results in `true`, `in` will evaluate to true; `false` otherwise.
 
--- a/docs/authoring-json-rules.md
+++ b/docs/authoring-json-rules.md
@@ -225,7 +225,7 @@ Example:
 #### **In**
 *Type: array of basic JSON values (integer, float, string, bool, null)*
 
-Evaluates the value of the `path` in the template using the `equals` operator for each value specified in the array.  If any results in `true`, `in` will evaluate to true; `false` otherwise.  All values in the array must be of the same type.
+Evaluates the value of the `path` in the template using the `equals` operator for each value specified in the array.  If any results in `true`, `in` will evaluate to true; `false` otherwise.
 
 Example:
 ```javascript
--- a/src/Analyzer.JsonRuleEngine.UnitTests/ExpressionConverterTests.cs
+++ b/src/Analyzer.JsonRuleEngine.UnitTests/ExpressionConverterTests.cs
@@ -33,6 +33,7 @@ namespace Microsoft.Azure.Templates.Analyzer.RuleEngines.JsonEngine.UnitTests
         [DataRow("equals", "someString", "exists", true, DisplayName = "{\"Equals\": \"someString\"}, Where: {\"Exists\": true}")]
         [DataRow("notEquals", 0, DisplayName = "{\"NotEquals\": 0}")]
         [DataRow("regex", "regexPattern", DisplayName = "{\"Regex\": \"regexPattern\"}")]
+        [DataRow("in", "aValue", DisplayName = "{\"In\": [\"anotherValue\", \"aValue\"]}")]
         public void ReadJson_LeafWithValidOperator_ReturnsCorrectTypeAndValues(string operatorProperty, object operatorValue, params object[] whereCondition)
         {
             // If whereCondition is populated, add a Where condition into JSON to parse.
--- a/src/Analyzer.JsonRuleEngine.UnitTests/ExpressionConverterTests.cs
+++ b/src/Analyzer.JsonRuleEngine.UnitTests/ExpressionConverterTests.cs
@@ -214,7 +214,10 @@ namespace Microsoft.Azure.Templates.Analyzer.RuleEngines.JsonEngine.UnitTests
         }
 
         [DataTestMethod]
+        [DataRow("in", 7, DisplayName = "\"In\": \"aString\"")]
+        [DataRow("in", 5.7, DisplayName = "\"In\": \"aString\"")]
         [DataRow("in", "aString", DisplayName = "\"In\": \"aString\"")]
+        [DataRow("in", true, DisplayName = "\"In\": \"aString\"")]
         [ExpectedException(typeof(JsonSerializationException))]
         public void ReadJson_LeafWithInvalidOperator_ThrowsSerializationException(string operatorProperty, object operatorValue)
         {
--- a/src/Analyzer.JsonRuleEngine.UnitTests/ExpressionConverterTests.cs
+++ b/src/Analyzer.JsonRuleEngine.UnitTests/ExpressionConverterTests.cs
@@ -25,6 +25,13 @@ namespace Microsoft.Azure.Templates.Analyzer.RuleEngines.JsonEngine.UnitTests
             .Where(property => property.Attribute != null)
             .ToDictionary(property => property.Attribute.PropertyName ?? property.Property.Name, property => property.Property, StringComparer.OrdinalIgnoreCase);
 
+        public static IReadOnlyList<object[]> InOperatorTestScenarios { get; } = new List<object[]>
+        {
+            new object[] { "in", new object[] { "anotherValue", "aValue"} }
+        }.AsReadOnly();
+
+        public static string GetInOperatorTestDisplayName(MethodInfo _, object[] data) => "{\"In\": [\"anotherValue\", \"aValue\"]}";
+
         [DataTestMethod]
         [DataRow("hasValue", true, DisplayName = "{\"HasValue\": true}")]
         [DataRow("hasValue", false, "hasValue", true, DisplayName = "{\"HasValue\": false}, Where: {\"HasValue\": true}")]
@@ -33,7 +40,7 @@ namespace Microsoft.Azure.Templates.Analyzer.RuleEngines.JsonEngine.UnitTests
         [DataRow("equals", "someString", "exists", true, DisplayName = "{\"Equals\": \"someString\"}, Where: {\"Exists\": true}")]
         [DataRow("notEquals", 0, DisplayName = "{\"NotEquals\": 0}")]
         [DataRow("regex", "regexPattern", DisplayName = "{\"Regex\": \"regexPattern\"}")]
-        [DataRow("in", "aValue", DisplayName = "{\"In\": [\"anotherValue\", \"aValue\"]}")]
+        [DynamicData(nameof(InOperatorTestScenarios), DynamicDataDisplayName = nameof(GetInOperatorTestDisplayName))]
         public void ReadJson_LeafWithValidOperator_ReturnsCorrectTypeAndValues(string operatorProperty, object operatorValue, params object[] whereCondition)
         {
             // If whereCondition is populated, add a Where condition into JSON to parse.
@@ -84,7 +91,8 @@ namespace Microsoft.Azure.Templates.Analyzer.RuleEngines.JsonEngine.UnitTests
                         }
                         else
                         {
-                            Assert.AreEqual(new JValue(value), parsedValue);
+                            var valueAsJson = JsonRuleEngineTestsUtilities.ToJToken(value);
+                            Assert.IsTrue(JToken.DeepEquals(valueAsJson, (JToken)parsedValue));
                         }
                     }
                     else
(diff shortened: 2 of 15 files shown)


docs/authoring-json-rules.md
```diff
@@ -222,10 +222,10 @@ Example:
 }
 ```
 
-#### **In [[*]](#note)**
+#### **In**
 *Type: array of basic JSON values (integer, float, string, bool, null)*
 
-Evaluates the value of the `path` in the template using the `equals` operator for each value specified in the array.  If any results in `true`, `in` will evaluate to true; `false` otherwise.  All values in the array must be of the same type.
+Evaluates the value of the `path` in the template using the `equals` operator for each value specified in the array.  If any results in `true`, `in` will evaluate to true; `false` otherwise.
```
JohnathonMohr, line 228: It would be helpful to add a small statement that different types within the array *are* allowed.

src/Analyzer.JsonRuleEngine.UnitTests/JsonRuleEngineTestsUtilities.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.Templates.Analyzer.RuleEngines.JsonEngine.UnitTests
+{
+    public class JsonRuleEngineTestsUtilities
```
JohnathonMohr, line 9: Rename class and file to TestUtilities.

src/Analyzer.JsonRuleEngine/Operators/InOperator.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// <summary>
+        /// Gets the name of this operator.
+        /// </summary>
+        public override string Name => "In";
+
+        /// <summary>
+        /// Creates an InOperator.
+        /// </summary>
+        /// <param name="specifiedValue">The value specified in the JSON rule, it has to be an array of basic JSON values.</param>
+        /// <param name="isNegative">Whether the result of EvaluateExpression() should be negated or not.</param>
```
JohnathonMohr, line 22: For references to code (like EvaluateExpression() here), put it in a `<see cref="..."/>` tag.

src/Analyzer.JsonRuleEngine/Operators/InOperator.cs
```diff
@@ -0,0 +1,50 @@
...
+        public InOperator(JToken specifiedValue, bool isNegative)
+        {
+            this.SpecifiedValue = specifiedValue;
```
JohnathonMohr, line 25: There should be validation somewhere that what's passed in is actually a JArray.  The earlier this validation can happen in the runtime, the better.  If it can be enforced as part of the schema, that would be best.

src/Analyzer.JsonRuleEngine/Operators/InOperator.cs
```diff
@@ -0,0 +1,50 @@
...
+            foreach (var arrayElement in this.SpecifiedValue)
+            {
+                if (equalsOperator.EvaluateExpression(arrayElement))
+                {
+                    return true;
+                }
+            }
+
+            return false;
+
```
JohnathonMohr, line 47: Remove extra line

src/Analyzer.JsonRuleEngine.UnitTests/InOperatorTests.cs
```diff
@@ -0,0 +1,36 @@
...
+    {
+        [DataTestMethod]
+        [DataRow(true, new object[] { "anotherValue", "aValue" }, DisplayName = "String is in string array")]
+        [DataRow(false, new object[] { "anotherValue", "otherValue" }, DisplayName = "String is not in string array")]
+        [DataRow(false, new object[] { }, DisplayName = "String is not in empty array")]
+        [DataRow(true, new object[] { "anotherValue", 4, false, 3.5, null, "aValue" }, DisplayName = "String is in mixed array")]
+        [DataRow(false, new object[] { "anotherValue", 4, false, 3.5, null, "otherValue" }, DisplayName = "String is not in mixed array")]
+        public void EvaluateExpression(bool evaluationResult, params object[] arrayOfValues)
+        {
+            var aValue = "aValue";
```
JohnathonMohr, line 21: Add tests to find other data types inside the array as well (besides a string).

src/Analyzer.JsonRuleEngine.UnitTests/InOperatorTests.cs
```diff
@@ -0,0 +1,36 @@
...
+            var inOperator = new InOperator(arrayOfValuesJToken, isNegative: false);
+            Assert.AreEqual(evaluationResult, inOperator.EvaluateExpression(valueJToken));
+        }
+
+        [TestMethod]
+        public void GetName_ReturnsCorrectName()
+        {
+            Assert.AreEqual("In", new InOperator(new JObject(), false).Name);
+        }
+    }
```
JohnathonMohr, line 35: Add tests for the constructor as well.

src/Analyzer.JsonRuleEngine.UnitTests/ExpressionConverterTests.cs
```diff
@@ -25,6 +25,13 @@ public class ExpressionConverterTests
             .Where(property => property.Attribute != null)
             .ToDictionary(property => property.Attribute.PropertyName ?? property.Property.Name, property => property.Property, StringComparer.OrdinalIgnoreCase);
 
+        public static IReadOnlyList<object[]> InOperatorTestScenarios { get; } = new List<object[]>
```
JohnathonMohr, line 28: Add a comment above these to explain they are for the ReadJson_LeafWithValidOperator_ReturnsCorrectTypeAndValues test.

JohnathonMohr (changes requested):
Looking pretty good overall.  Just a few things to add.

src/Analyzer.JsonRuleEngine/Operators/InOperator.cs
VeraBE: Adding it everywhere then :)

src/Analyzer.JsonRuleEngine.UnitTests/InOperatorTests.cs
VeraBE: I thought it wasn't necessary because for the comparison I'm using the 'equals' operator that covers those cases in its own tests, do you think I should add those tests anyways?

JohnathonMohr: Using the 'equals' operator is a good way to implement it.  Having tests ensures it's working as intended, and also protects against future changes that may inadvertently break it.

src/Analyzer.JsonRuleEngine/Operators/InOperator.cs
VeraBE: I'm thinking on adding it in LeafExpressionDefinition, I couldn't find an earlier stage to make the validation, what do you think?

src/Analyzer.JsonRuleEngine.UnitTests/InOperatorTests.cs
VeraBE: I'm assuming you wrote this because of the array validation issue. I added a test for that in ExpressionConverterTests. It doesn't even compile if I try to use a non JArray as parameter in the constructor

src/Analyzer.JsonRuleEngine.UnitTests/ExpressionConverterTests.cs
```diff
@@ -210,12 +210,15 @@ void ValidateExpression(ExpressionDefinition expression, Type expectedSpecificTy
...
-                {{
-                    ""resourceType"": ""resourceType"",
-                    ""path"": ""path"",
-                    ""{operatorProperty}"": {JsonConvert.SerializeObject(operatorValue)}
-                }}");
+            ReadSimpleOperatorJson(operatorProperty, operatorValue);
+        }
+
+        [DataTestMethod]
+        [DataRow("in", "aString", DisplayName = "\"In\": \"aString\"")]
```
JohnathonMohr, line 217: Add tests for other types as well.

src/Analyzer.JsonRuleEngine.UnitTests/InOperatorTests.cs
```diff
@@ -39,7 +38,7 @@ public void EvaluateExpression(bool evaluationResult, object desiredValue, param
         [TestMethod]
         public void GetName_ReturnsCorrectName()
         {
-            Assert.AreEqual("In", new InOperator(new JObject(), false).Name);
+            Assert.AreEqual("In", new InOperator(new JArray(), false).Name);
         }
     }
```
JohnathonMohr, line 43: Also add a test for evaluating a null JToken.  This is different than evaluating a JToken for which the value in it is null.
For example, something like this:
``` diff
- var desiredValueJToken = TestUtilities.ToJToken(null);
+ var desiredValueJToken = null;
inOperator.EvaluateExpression(desiredValueJToken)
```

This will probably need to be a test on its own, since it doesn't fit what other tests are doing.

JohnathonMohr (changes requested):

src/Analyzer.JsonRuleEngine.UnitTests/InOperatorTests.cs
VeraBE: I can't compile `inOperator.EvaluateExpresion(null)`: _Cannot assign <null> to an implicitly typed variable_

JohnathonMohr: Whoops - `JToken desiredValueJToken = null;` 🙂

src/Analyzer.JsonRuleEngine/Operators/InOperator.cs
```diff
@@ -0,0 +1,54 @@
...
+        /// <param name="tokenToEvaluate">The JToken to evaluate.</param>
+        /// <returns>A value indicating whether or not the evaluation passed.</returns>
+        public override bool EvaluateExpression(JToken tokenToEvaluate)
+        {
+            if (tokenToEvaluate == null)
+            {
+                return false;
+            }
+
+            var equalsOperator = new EqualsOperator(tokenToEvaluate, isNegative: false);
```
yane3628, line 41: I believe this should pass the value `this.IsNegative` instead of `false`.

yane3628 (changes requested):

src/Analyzer.JsonRuleEngine/Operators/InOperator.cs
VeraBE: This operator doesn't use the isNegative attribute, it's always false, do you think I should change it anyways?

src/Analyzer.JsonRuleEngine.UnitTests/InOperatorTests.cs
```diff
@@ -0,0 +1,55 @@
...
+        [DataRow(true, false, new object[] { 4, 3.5, "aValue", false, null }, DisplayName = "Boolean is in")]
+        [DataRow(false, true, new object[] { 4, 3.5, "aValue", false, null }, DisplayName = "Boolean is not in")]
+        [DataRow(true, null, new object[] { 4, 3.5, "aValue", false, null }, DisplayName = "Null is in")]
+        [DataRow(false, null, new object[] { 4, 3.5, "aValue", false }, DisplayName = "Null is not in")]
+        [DataRow(false, 7, new object[] { }, DisplayName = "Integer is not in empty array")]
+        [DataRow(false, 7.7, new object[] { }, DisplayName = "Float is not in empty array")]
+        [DataRow(false, "aValue", new object[] { }, DisplayName = "String is not in empty array")]
+        [DataRow(false, true, new object[] { }, DisplayName = "Boolean is not in empty array")]
+        [DataRow(false, null, new object[] { }, DisplayName = "Null is not in empty array")]
+        public void EvaluateExpression(bool evaluationResult, object desiredValue, object arrayOfValues)
```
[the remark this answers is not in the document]
VeraBE: I see how I could do that for the null test, something like: EvaluateExpression_NullAsTokenToEvaluate_ReturnsFalse, right? But for the other test the outcome and the case being tested changes in each DataRow. It's okay if I leave that one the way it is?

src/Analyzer.JsonRuleEngine/Operators/InOperator.cs
JohnathonMohr: This is what I mentioned offline.  You are setting `this.IsNegative` in the constructor, but aren't using it here.  Either don't expose it as a parameter in the constructor, or honor whatever it's set to and add tests for it.

src/Analyzer.JsonRuleEngine.UnitTests/InOperatorTests.cs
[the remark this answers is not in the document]
yane3628: You can do it in a generic way. Something like EvaluateExpression_ValidDataType_ReturnsExpectedEvaluationResult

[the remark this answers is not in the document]
JohnathonMohr: The Equals operator tests have good examples, like `EvaluateExpression_PropertyIsEqual_EqualsExpressionIsTrue_NotEqualsExpressionIsFalse`.  It might be good to separate this test into two - one where the value is in the array, one where it isn't.

JohnathonMohr (approved):

yane3628 (approved):

PR #102, opened 2021-05-04
Add metadata strings for operators and improve CLI output

### LINKED ISSUE #33: Add metadata to result to help identify exactly what part of the check was violated
- Add message for determining the error
- Report Line Number and other metadata to CLI results

In an AllOf expression, this will help identify which sub-expressions caused the AllOf to fail.

Related Issue: https://github.com/Azure/template-analyzer/issues/199

yane3628:
- Add failure message for each operator (Closes #33)
- Add tests for failure message generation
- Add line number, file path, and failure message to CLI output


yane3628:
Structured expression messages still need to be implemented.

src/Analyzer.JsonRuleEngine/JsonRuleResult.cs
```diff
@@ -27,8 +29,32 @@ internal class JsonRuleResult : IResult
...
+
+        /// <summary>
+        /// Gets the actual value present at the specified path.
+        /// </summary>
+        internal JToken ActualValue { get; set; }
+
+        /// <summary>
+        /// Gets the messsage which explains why the evaluation failed.
+        /// </summary>
+        public string FailureMessage()
```
JohnathonMohr, line 44: As much as possible, I think all this logic should be in the Operators and Expressions.  I think it would be easier to understand and write if the failure messages were built right where the evaluations happen.

JohnathonMohr (changes requested):
I think changing the design to have messages finalized inside the Expression and Operator (when the Result is populated) will help the code flow more smoothly.  I'm also thinking about what the messaging for the `not` operator would look like.  I'm not sure how best to represent that yet, but something we need to decide before committing to a design here.

PR #105, opened 2021-05-04
Add line number, file path, and parameters path to CLI output

yane3628:
Add line number, file path, and parameters path to CLI output

### DIFF
--- a/src/Analyzer.Cli/CommandLineParser.cs
+++ b/src/Analyzer.Cli/CommandLineParser.cs
@@ -14,7 +14,7 @@ namespace Microsoft.Azure.Templates.Analyzer.Cli
     {
         RootCommand rootCommand;
         private readonly string IndentedNewLine = Environment.NewLine + "\t";
-        private readonly string TwiceIndentedNewLine = Environment.NewLine + "\t";
+        private readonly string TwiceIndentedNewLine = Environment.NewLine + "\t\t";
 
         /// <summary>
         /// Constructor for the command line parser. Sets up the command line API. 
--- a/src/Analyzer.Cli/CommandLineParser.cs
+++ b/src/Analyzer.Cli/CommandLineParser.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.CommandLine;
 using System.CommandLine.Invocation;
 using System.IO;
+using System.Linq;
 using System.Threading.Tasks;
 
 namespace Microsoft.Azure.Templates.Analyzer.Cli
@@ -80,7 +81,8 @@ namespace Microsoft.Azure.Templates.Analyzer.Cli
 
                     foreach (var evaluation in evaluations)
                     {
-                        Console.WriteLine($"{evaluation.RuleName}: {evaluation.RuleDescription}, Result: {evaluation.Passed.ToString()}");
+                        string resultString = GenerateResultString(evaluation, parametersFilePath == null ? null : File.ReadAllText(parametersFilePath.FullName));
+                        Console.WriteLine($"\n\n{evaluation.RuleName}: {evaluation.RuleDescription}\n\tFile: {templateFilePath}\n\tResult: {evaluation.Passed} {resultString}");
                     }
                 }
                 catch (Exception exp)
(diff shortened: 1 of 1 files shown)


src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -93,6 +100,31 @@ private Command SetupAnalyzeTemplateCommand()
             return analyzeTemplateCommand;
         }
 
+        private string GenerateResultString(Types.IEvaluation evaluation)
+        {
+            string resultString = "";
+
+            if (!evaluation.Passed)
+            {
+                if (evaluation.Results.Count() > 0)
```
JohnathonMohr, line 109: This won't display properly with evaluations that have results in addition to sub-evaluations with results.  I suggest relying on `evaluation.HasResults` when looking at each evaluation.

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -80,7 +81,13 @@ private Command SetupAnalyzeTemplateCommand()
 
                     foreach (var evaluation in evaluations)
                     {
-                        Console.WriteLine($"{evaluation.RuleName}: {evaluation.RuleDescription}, Result: {evaluation.Passed.ToString()}");
+                        string resultString = GenerateResultString(evaluation);
+                        string fileMetadata = $"\n\tFile: {templateFilePath}";
```
JohnathonMohr, line 85: template and parameters will be the same for every evaluation.  Can that be printed just once?

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -80,7 +81,13 @@ private Command SetupAnalyzeTemplateCommand()
...
+                        if (parametersFilePath != null)
+                        {
+                            fileMetadata += $"\n\tParameters File: {parametersFilePath}";
+                        };
+                        Console.WriteLine($"\n\n{evaluation.RuleName}: {evaluation.RuleDescription}{fileMetadata}\n\tResult: {evaluation.Passed} {resultString}");
```
JohnathonMohr, line 90: Instead of printing true/false for the result, can we do Pass/Fail instead?

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -93,6 +100,31 @@ private Command SetupAnalyzeTemplateCommand()
...
+        {
+            string resultString = "";
+
+            if (!evaluation.Passed)
+            {
+                if (evaluation.Results.Count() > 0)
+                {
+                    foreach (var result in evaluation.Results)
+                    {
+                        resultString += $"\n\tLine: {result.LineNumber}";
```
JohnathonMohr, line 113: `\n` by itself may not result in a proper new line in some Windows prompts or text viewers.  Use `Environment.NewLine` instead.

It would probably be helpful to create a constant in this file, something like `const string IndentedNewLine = Environment.NewLine + "\t";`

JohnathonMohr (changes requested):

src/Analyzer.Cli/CommandLineParser.cs
yane3628: Can you give an example of this? I tested it with HasResults, but for the case of allOf with two leaf expressions HasResult was true. I will just remove the check for if there is Results instead. If the count is 0, it won't loop through the for loop anyway.

src/Analyzer.Cli/CommandLineParser.cs
JohnathonMohr: I meant only print the template and parameters file names once for everything.  Every evaluation here is for the same template, so there's no need to keep reprinting it.

src/Analyzer.Cli/CommandLineParser.cs
JohnathonMohr: I don't understand what you mean - should allOf with two leafs *not* return true for HasResults?

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -13,6 +13,8 @@ namespace Microsoft.Azure.Templates.Analyzer.Cli
     internal class CommandLineParser
     {
         RootCommand rootCommand;
+        private readonly string IndentedNewLine = Environment.NewLine + "\t";
+        private readonly string TwiceIndentedNewLine = Environment.NewLine + "\t";
```
JohnathonMohr, line 17: Add another "\t"?

src/Analyzer.Cli/CommandLineParser.cs
yane3628: I was trying to use HasResults to know if it's a leafexpression or a structured, but it that's not what it was intended for.

src/Analyzer.Cli/CommandLineParser.cs
yane3628: I moved it and now it should only be printing once.

PR #107, opened 2021-05-04
Add additional built in rules

### LINKED ISSUE #43: [Built-in Rule] Only secure connections to your Azure Cache for Redis should be enabled
Azure policy: https://github.com/Azure/azure-policy/blob/master/built-in-policies/policyDefinitions/Cache/RedisCache_AuditSSLPort_Audit.json

``` javascript
{
    "name": "Cache_Redis_SSL_Only",
    "description": "Only secure connections to your Azure Cache for Redis should be enabled",
    "recommendation": "Enable connections via SSL only to Redis Cache",
    "helpUri": "https://github.com/Azure/template-analyzer/docs/built-in-bpa-rules.md/#only-secure-connections-to-your-azure-cache-for redis-should-be-enabled",
    "evaluation": {
      "resourceType": "Microsoft.Cache/redis",
      "path": "properties.enableNonSslPort",
      "equals": false
    }
}
```

### LINKED ISSUE #83: [Built-in Rule] Ensure all virtual machines are not using preview images
Original logic: https://github.com/Azure/arm-ttk/blob/master/arm-ttk/testcases/deploymentTemplate/Virtual-Machines-Should-Not-Be-Preview.test.ps1

``` javascript
{
    "name": "Compute_VirtualMachine_ShouldNotBePreview",
    "description": "Ensure all virtual machines are not using preview images",
    "recommendation": "Use a non-preview image for virtual machines",
    "helpUri": "https://github.com/Azure/template-analyzer/docs/built-in-bpa-rules.md/#ensure-all-virtual-machines-are-not-using-preview-images",
    "evaluation": {
      "anyOf": [
        {
          "path": "properties.storageProfile.imageReference",
          "regex": "^((?!-preview$).)*$"
        },
        {
          "path": "properties.storageProfile.imageReference.version",
          "regex": "^((?!-preview$).)*$"
        }
      ]
    }
  },
{
    "name": "Compute_VirtualMachineScaleSets_ShouldNotBePreview",
    "description": "Ensure all virtual machine scale set instances are not using preview images",
    "recommendation": "Use a non-preview image for virtual machine scale set instances",
    "helpUri": "https://github.com/Azure/template-analyzer/docs/built-in-bpa-rules.md/#ensure-all-virtual-machine-scale-set-instances-are-not-using-preview-images",
    "evaluation": {
      "resourceType": "Microsoft.Compute/virtualMachineScaleSets",
      "anyOf": [
        {
          "path": "properties.virtualMachineProfile.storageProfile.imageReference",
          "regex": "^((?!-preview$).)*$"
        },
        {
          "path": "properties.virtualMachineProfile.storageProfile.imageReference.version",
          "regex": "^((?!-preview$).)*$"
        }
      ]
    }
  }
```

yane3628:
- Add rule: Only secure connections to your Azure Cache for Redis should be enabled (Closes #43)
- Add rule: Ensure all virtual machines are not using preview images (Closes #83)


src/Analyzer.Core/Rules/BuiltInRules.json
```diff
@@ -86,6 +86,55 @@
...
+    }
+  },
+  {
+    "name": "Compute_VirtualMachineScaleSets_ShouldNotBePreview",
+    "description": "Ensure all virtual machine scale set instances are not using preview images",
+    "recommendation": "Use a non-preview image for virtual machine scale set instances",
+    "helpUri": "https://github.com/Azure/template-analyzer/docs/built-in-bpa-rules.md/#ensure-all-virtual-machine-scale-set-instances-are-not-using-preview-images",
+    "evaluation": {
+      "resourceType": "Microsoft.Compute/virtualMachineScaleSets",
+      "anyOf": [
```
yane3628, line 126: @JohnathonMohr's comment from previous PR

I'm fairly confident the TTK is incorrectly checking this property. But fortunately I don't think there's any harm keeping this format for now.

docs/built-in-bpa-rules.md
```diff
@@ -19,6 +30,11 @@ Function Apps should require HTTPS to ensure connections are made to the expecte
...
 
+## Only secure connections to your Azure Cache for Redis should be enabled
+Enable only connections via SSL to Redis Cache. Use of secure connections ensures authentication between the server and the service and protects data in transit from network layer attacks such as man-in-the-middle, eavesdropping, and session-hijacking.
+
+**Recommendation**: To [enable only connections via SSL to Redis Cache](https://docs.microsoft.com/en-us/security/benchmark/azure/baselines/azure-cache-for-redis-security-baseline?toc=/azure/azure-cache-for-redis/TOC.json#44-encrypt-all-sensitive-information-in-transit), in the [Microsoft.Cache/Redis resource properties](https://docs.microsoft.com/en-us/azure/templates/microsoft.cache/redis?tabs=json#rediscreateproperties-object), add (or update) the *enableNonSslPort* property, setting its value to `false`.
```
JohnathonMohr, line 36: We should update the guidance slightly, since we're taking the default value of 'false' into account.  They can change the value from true to false, or just delete the property.

JohnathonMohr (approved):
The Redis Cache rule needs a slight change in recommendation, but looks good overall.

docs/built-in-bpa-rules.md
yane3628: Thoughts on this wording?

Recommendation: To enable only connections via SSL to Redis Cache, in the Microsoft.Cache/Redis resource properties, add (or update) the enableNonSslPort property, setting its value to false or remove the property if the Microsoft.Cache/redis Template API version you are using has false as it's default value.

JohnathonMohr: I don't think we need to worry about different API versions in the recommendation.  It's extremely unlikely a new version will be published that changes the default to 'true', making it less secure than previous versions.  However, if *previous* versions have a different default value, we should change the rule back to enforce false.

Currently, since the rule only requires the property isn't 'true', the recommendation shouldn't suggest adding the property, since it won't flag to begin with if the property doesn't already exist.  Another suggestion:

Recommendation: To enable only connections via SSL to Redis Cache, in the Microsoft.Cache/Redis resource properties, update the value of the *enableNonSslPort* property from `true` to `false`, or remove the property from the template.

yane3628: My concern is if previous versions had a different default value. I'll update the recommendation, but we should test out previous versions.

yane3628: I just made test deployments. Default value for versions "2020-06-01" and later have false as the default value.

src/Analyzer.Core/Rules/BuiltInRules.json
yane3628: We are still pending confirmation if this rule is 100% correct.

PR #116, opened 2021-05-27
Support writing rules that depend on the relationship between resources

### LINKED ISSUE #71: Template Expansion: Put dependent resources into dependency resource
In an ARM template, a resource can be designated as depending upon other resources in the template, like:
``` javascript
"resources": [
  {
    "name": "<dependency-name>"
    "type": "<dependency-type>",
    ...
  },
  {
    "name": "<name-of-dependent-resource>"
    "type": "<some-resource-type>",
    "dependsOn": [
      "<dependency-type>/<dependency-name>",
      ...
    ]
    ...
  }
]
```
When processing and expanding a template, when these resources are found, a copy of them should be put into the resources sub-array of each resource they depend on.

Then in the expanded template, the dependency resource would look something like:
``` diff
{
  "name": "<dependency-name>"
  "type": "<dependency-type>",
  ...,
+  "resources": [
+    {
+      "name": "<name-of-dependent-resource>"
+      "type": "<some-resource-type>",
+      ...
+    }
+  ]
}
```

This will enable rules to be written where the rule only applies to a resource type if some condition about its parent is met.

### LINKED ISSUE #94: [Built-in Rule] Latest TLS version should be used in your Web App
Azure Policy link: https://github.com/Azure/azure-policy/blob/master/built-in-policies/policyDefinitions/App%20Service/AppService_RequireLatestTls_WebApp_Audit.json

``` js
{
    "name": "AppServiceWebApp_RequireLatestTls",
    "description": "Latest TLS version should be used in your Web App",
    "recommendation": "Upgrade to the latest TLS version",
    "helpUri": "https://github.com/Azure/template-analyzer/docs/built-in-bpa-rules.md/#latest-tls-version-should-be-used-in-your-web-app",
    "evaluation": {
      "resourceType": "Microsoft.Web/sites",
      "where": {
          "anyOf": [
          {
            "path": "kind",
            "exists": false
          },
          {
            "path": "kind",
            "regex": "^app"
          }
        ]
      },
      "evaluate": {
        "resourceType": "Microsoft.Web/sites/config",
        "path": "properties.minTlsVersion",
        "equals": "1.2"
      }
    }
  }
```

yane3628:
- Map resources that depend on other resources to support rules that use the where clause (Closes #71)
- Report the correct line number when mapping the result to original template
- Add tests
- Add built-in rule "Latest TLS version should be used in your Web App" (Closes #94)

### DIFF
--- a/docs/built-in-bpa-rules.md
+++ b/docs/built-in-bpa-rules.md
@@ -30,6 +30,12 @@ Function Apps should require HTTPS to ensure connections are made to the expecte
 
 **Recommendation**: To [use HTTPS to ensure server/service authentication and protect data in transit from network layer eavesdropping attacks](https://docs.microsoft.com/en-us/azure/app-service/configure-ssl-bindings#enforce-https), in the [Microsoft.Web/Sites resource properties](https://docs.microsoft.com/en-us/azure/templates/microsoft.web/sites?tabs=json#siteproperties-object), add (or update) the *httpsOnly* property, setting its value to `true`.
 
+## Latest TLS version should be used in your Web App
+Web Apps should require the latest TLS version.
+
+**Recommendation**: 
+To [enforce the latest TLS version](https://docs.microsoft.com/en-us/azure/app-service/configure-ssl-bindings#enforce-tls-versions), in the [Microsoft.Web/sites/config resource properties](https://docs.microsoft.com/en-us/azure/templates/microsoft.web/sites/config-web?tabs=json#SiteConfig), add (or update) the *minTlsVersion* property, setting its value to `1.2`.
+
 ## Only secure connections to your Azure Cache for Redis should be enabled
 Enable only connections via SSL to Redis Cache. Use of secure connections ensures authentication between the server and the service and protects data in transit from network layer attacks such as man-in-the-middle, eavesdropping, and session-hijacking.
 
--- a/docs/built-in-bpa-rules.md
+++ b/docs/built-in-bpa-rules.md
@@ -1,17 +1,17 @@
 # Best Practice Related Rules
-## Ensure all virtual machines are not using preview images
-Virtual Machines should not use preview versions of images.
+## Ensure all Virtual Machines are not using preview images
+Virtual machines should not use preview versions of images.
 
 **Recommendation**: To use a non-preview image for virtual machines, in the [Microsoft.Compute/VirtualMachines/imageReference resource properties](https://docs.microsoft.com/en-us/azure/templates/microsoft.compute/virtualmachines?tabs=json#ImageReference), ensure the *version* property does not contain "-preview".
 
-## Ensure all virtual machine scale set instances are not using preview images
-Virtual Machine Scale Set instances should not use preview versions of images.
+## Ensure all Virtual Machine Scale Set instances are not using preview images
+Virtual machine scale set instances should not use preview versions of images.
 
 **Recommendation**: To use a non-preview image for virtual machine scale set instances, in the [Microsoft.Compute/VirtualMachineScaleSets/storageProfile.imageReference resource properties](https://docs.microsoft.com/en-us/azure/templates/microsoft.compute/virtualmachinescalesets?tabs=json#ImageReference), ensure the *version* property does not contain "-preview".
 
 # Security Related Rules
 ## API App Should Only Be Accessible Over HTTPS
-API Apps should require HTTPS to ensure connections are made to the expected server and data in transit is protected from network layer eavesdropping attacks.
+API apps should require HTTPS to ensure connections are made to the expected server and data in transit is protected from network layer eavesdropping attacks.
 
 **Recommendation**: To [use HTTPS to ensure server/service authentication and protect data in transit from network layer eavesdropping attacks](https://docs.microsoft.com/en-us/azure/app-service/configure-ssl-bindings#enforce-https), in the [Microsoft.Web/Sites resource properties](https://docs.microsoft.com/en-us/azure/templates/microsoft.web/sites?tabs=json#siteproperties-object), add (or update) the *httpsOnly* property, setting its value to `true`.
 
@@ -26,12 +26,12 @@ It is important to enable encryption of Automation account variable assets when 
 **Recommendation**: [Enable encryption of Automation account variable assets](https://docs.microsoft.com/en-us/azure/automation/shared-resources/variables?tabs=azure-powershell)
 
 ## Function App Should Only Be Accessible Over HTTPS
-Function Apps should require HTTPS to ensure connections are made to the expected server and data in transit is protected from network layer eavesdropping attacks.
+Function apps should require HTTPS to ensure connections are made to the expected server and data in transit is protected from network layer eavesdropping attacks.
 
 **Recommendation**: To [use HTTPS to ensure server/service authentication and protect data in transit from network layer eavesdropping attacks](https://docs.microsoft.com/en-us/azure/app-service/configure-ssl-bindings#enforce-https), in the [Microsoft.Web/Sites resource properties](https://docs.microsoft.com/en-us/azure/templates/microsoft.web/sites?tabs=json#siteproperties-object), add (or update) the *httpsOnly* property, setting its value to `true`.
 
 ## Latest TLS version should be used in your Web App
-Web Apps should require the latest TLS version.
+Web apps should require the latest TLS version.
 
 **Recommendation**: 
 To [enforce the latest TLS version](https://docs.microsoft.com/en-us/azure/app-service/configure-ssl-bindings#enforce-tls-versions), in the [Microsoft.Web/sites/config resource properties](https://docs.microsoft.com/en-us/azure/templates/microsoft.web/sites/config-web?tabs=json#SiteConfig), add (or update) the *minTlsVersion* property, setting its value to `1.2`.
@@ -57,6 +57,6 @@ You should only use built-in roles instead of cutom RBAC roles. Custom RBAC role
 **Recommendation**: [Use built-in roles such as 'Owner, Contributer, Reader' instead of custom RBAC roles](https://docs.microsoft.com/en-us/azure/role-based-access-control/built-in-roles)
 
 ## Web Application Should Only Be Accessible Over HTTPS
-Web Apps should require HTTPS to ensure connections are made to the expected server and data in transit is protected from network layer eavesdropping attacks.
+Web apps should require HTTPS to ensure connections are made to the expected server and data in transit is protected from network layer eavesdropping attacks.
 
 **Recommendation**: To [use HTTPS to ensure server/service authentication and protect data in transit from network layer eavesdropping attacks](https://docs.microsoft.com/en-us/azure/app-service/configure-ssl-bindings#enforce-https), in the [Microsoft.Web/Sites resource properties](https://docs.microsoft.com/en-us/azure/templates/microsoft.web/sites?tabs=json#siteproperties-object), add (or update) the *httpsOnly* property, setting its value to `true`.
--- a/src/Analyzer.Core.UnitTests/TemplateAnalyzerTests.cs
+++ b/src/Analyzer.Core.UnitTests/TemplateAnalyzerTests.cs
@@ -18,7 +18,7 @@ namespace Microsoft.Azure.Templates.Analyzer.Core.UnitTests
         public void EvaluateRulesAgainstTemplate_ValidInputValues_ReturnCorrectEvaluations(string resource1Properties, string resourceType, int expectedEvaluationCount, bool expectedEvaluationPassed, string resource2Properties = null)
         {
             // Arrange
-            string[] resourceProperties = { GenerateResource(resource1Properties, resourceType), GenerateResource(resource2Properties, resourceType) };
+            string[] resourceProperties = { GenerateResource(resource1Properties, resourceType, "resource1"), GenerateResource(resource2Properties, resourceType, "resource2") };
             string template = GenerateTemplate(resourceProperties);
 
             TemplateAnalyzer templateAnalyzer = new TemplateAnalyzer(template);
@@ -38,7 +38,7 @@ namespace Microsoft.Azure.Templates.Analyzer.Core.UnitTests
             ""resources"": [ {0} ] }}", string.Join(',', resourceProperties));
         }
 
-        private string GenerateResource(string resourceProperties, string resourceType)
+        private string GenerateResource(string resourceProperties, string resourceType, string resourceName)
         {
             if (string.IsNullOrEmpty(resourceProperties))
             {
@@ -48,10 +48,10 @@ namespace Microsoft.Azure.Templates.Analyzer.Core.UnitTests
             return string.Format(@"
                     {{
                       ""apiVersion"": ""2018-02-01"",
-                      ""name"": ""resourceName"",
+                      ""name"": ""{2}"",
                       ""type"": ""{1}"",
                       ""properties"": {0}
-                    }}", resourceProperties, resourceType);
+                    }}", resourceProperties, resourceType, resourceName);
         }
 
         [TestMethod]
--- a/src/Analyzer.Core/Rules/BuiltInRules.json
+++ b/src/Analyzer.Core/Rules/BuiltInRules.json
@@ -65,6 +65,34 @@
     }
   },
   {
+    "name": "AppServiceWebApp_RequireLatestTls",
+    "description": "Latest TLS version should be used in your Web App",
+    "recommendation": "Upgrade to the latest TLS version",
+    "helpUri": "https://github.com/Azure/template-analyzer/docs/built-in-bpa-rules.md/#latest-tls-version-should-be-used-in-your-web-app",
+    "evaluation": {
+      "resourceType": "Microsoft.Web/sites",
+      "where": {
+        "anyOf": [
+          {
+            "path": "kind",
+            "exists": false
+          },
+          {
+            "path": "kind",
+            "regex": "^app"
+          }
+        ]
+      },
+      "allOf": [
+        {
+          "resourceType": "Microsoft.Web/sites/config",
+          "path": "properties.minTlsVersion",
+          "equals": "1.2"
+        }
+      ]
+    }
+  },
+  {
     "name": "Authorization_RoleDefinition_No_Custom_RBAC",
     "description": "Use built-in roles instead of custom RBAC roles",
     "recommendation": "Use built-in roles such as 'Owner, Contributer, Reader' instead of custom RBAC roles",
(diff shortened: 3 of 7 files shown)


src/Analyzer.TemplateProcessor/ArmTemplateProcessor.cs
```diff
@@ -182,6 +173,133 @@ internal Template ProcessTemplateResourceAndOutputPropertiesLanguageExpressions(
...
+                    foreach (var childResource in templateResource.Resources)
+                    {
+                        ProcessTemplateResourceLanguageExpressions(childResource, evaluationHelper);
+                    }
+                }
+            }
+            catch
+            {
+                // Do not throw if there was an issue with evaluating language expressions
+                return null;
```
VeraBE, line 297: I don't see usages of the return values of this function

src/Analyzer.TemplateProcessor/ArmTemplateProcessor.cs
```diff
@@ -124,39 +125,29 @@ internal Template ParseAndValidateTemplate(InsensitiveDictionary<JToken> paramet
...
-                    }
-                }
-                catch
-                {
-                    // Do not throw if there was an issue with evaluating language expressions
-                    continue;
-                }
+                ProcessTemplateResourceLanguageExpressions(resource, evaluationHelper);
+
+                CopyResourceDependants(resource, flattenedResources);
```
VeraBE, line 150: If the call to ProcessTemplateResourceLanguageExpressions fails, won't CopyResourceDependant have problems and potentially throw an unexpected exception?

I think it could definitely happen if the return value of  ProcessTemplateResourceLanguageExpressions is meant to be used

docs/built-in-bpa-rules.md
```diff
@@ -30,6 +30,12 @@ Function Apps should require HTTPS to ensure connections are made to the expecte
 
 **Recommendation**: To [use HTTPS to ensure server/service authentication and protect data in transit from network layer eavesdropping attacks](https://docs.microsoft.com/en-us/azure/app-service/configure-ssl-bindings#enforce-https), in the [Microsoft.Web/Sites resource properties](https://docs.microsoft.com/en-us/azure/templates/microsoft.web/sites?tabs=json#siteproperties-object), add (or update) the *httpsOnly* property, setting its value to `true`.
 
+## Latest TLS version should be used in your Web App
+Web Apps should require the latest TLS version.
```
VeraBE, line 34: "web app" in lowecase?

src/Analyzer.Core.UnitTests/TemplateAnalyzerTests.cs
```diff
@@ -48,10 +48,10 @@ private string GenerateResource(string resourceProperties, string resourceType)
             return string.Format(@"
                     {{
                       ""apiVersion"": ""2018-02-01"",
-                      ""name"": ""resourceName"",
+                      ""name"": ""{2}"",
                       ""type"": ""{1}"",
                       ""properties"": {0}
-                    }}", resourceProperties, resourceType);
+                    }}", resourceProperties, resourceType, resourceName);
         }
```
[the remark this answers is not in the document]
yane3628: I believe it's because I do a match with resource name and type, I needed the names to be different. It's expecting name and resource type combinations to be unique in a template.

src/Analyzer.TemplateProcessor/ArmTemplateProcessor.cs
yane3628: True. I can make it void

src/Analyzer.TemplateProcessor/ArmTemplateProcessor.cs
yane3628: Since there we're catching all exceptions in ProcessTemplateResourceLanguageExpressions I don't think so... ProcessTemplateResourceLanguageExpressions doesn't need to return anything since  all the changes it makes are by reference anyway. I can't think of a case where ProcessTemplateResourceLanguageExpressions eats an exception there would be issues in CopyResourceDependants.

src/Analyzer.Core/Rules/BuiltInRules.json
```diff
@@ -64,6 +64,34 @@
       ]
     }
   },
+  {
+    "name": "AppServiceWebApp_RequireLatestTls",
+    "description": "Latest TLS version should be used in your Web App",
```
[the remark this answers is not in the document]
yane3628: I'm going to leave this one so that it matches the ASC recommendation, but I updated the description in the docs.

VeraBE (approved):

src/Analyzer.Utilities/JsonLineNumberResolver.cs
```diff
@@ -116,18 +118,52 @@ public int ResolveLineNumber(string pathInExpandedTemplate)
...
+                    && !originalTemplatePath.EndsWith("resources", StringComparison.OrdinalIgnoreCase)))
             {
                 return (tokenFromOriginalTemplate as IJsonLineInfo)?.LineNumber ?? 0;
             }
 
-            // The first unmatched property must be a resources[] array, likely a subresource
-            // of a top-level resource.  This is possible if resources were copied into other
-            // resources as part of template expansion.
-            // TODO
+            string[] pathSegments = unmatchedPathInOriginalTemplate.Split('.');
```
JohnathonMohr, line 132: Add a comment for what's going to happen next (since it didn't return above).

src/Analyzer.Utilities/JsonLineNumberResolver.cs
```diff
@@ -116,18 +118,52 @@ public int ResolveLineNumber(string pathInExpandedTemplate)
...
+            string remainingPathAtResourceScope = string.Join('.', pathSegments[1..]);
+
+            var match2 = childResourceIndexInPath.Match(pathInExpandedTemplate);
```
JohnathonMohr, line 135: Why not reuse the `match` variable?

JohnathonMohr, line 135: I think you need to parse out the resource that isn't found from `unmatchedPathInOriginalTemplate` instead of relying on a regex against the entire path.

For example, a resource could be copied into another child resource that is in the original template (so the resource is now a grandchild) and there's a problem in the grandchild.  In this example, the path would be *resources[x].resources[y].resources[z].somepath*, and the unmatched part would be *resources[z].somepath*.

Relying on the regex, *resources[x].resources[y]* would be found, and then the next call to `ResolveLineNumber` would use the incorrect path *resources[x].resources[y].somepath* (since *resources[z]* is removed from the unmatched part).

src/Analyzer.Utilities/JsonLineNumberResolver.cs
```diff
@@ -116,18 +118,52 @@ public int ResolveLineNumber(string pathInExpandedTemplate)
...
             // Compare the path of the expanded template with the path of the JToken found in the original template.
-            // If they match, or if the first unmatched property in the original is NOT a resources array,
+            // If they match or if the first unmatched property in the original is NOT a resources array
+            // and it's looking in a resources array and the first unmatched property in the original is an array index
             // the line number from the JToken of the original template can be returned directly.
             if (originalTemplatePath.Length == pathInExpandedTemplate.Length
-                || !unmatchedPathInOriginalTemplate.StartsWith("resources", StringComparison.OrdinalIgnoreCase))
+                || (!unmatchedPathInOriginalTemplate.StartsWith("resources", StringComparison.OrdinalIgnoreCase) 
+                    && !unmatchedPathInOriginalTemplate.StartsWith("[", StringComparison.OrdinalIgnoreCase)
+                    && !originalTemplatePath.EndsWith("resources", StringComparison.OrdinalIgnoreCase)))
```
JohnathonMohr, line 127: See if you can make this a little more readable/understandable.  It might even be worth splitting this up into multiple `if`s.

JohnathonMohr (changes requested):

docs/built-in-bpa-rules.md
```diff
@@ -1,17 +1,17 @@
...
-Virtual Machines should not use preview versions of images.
+## Ensure all Virtual Machines are not using preview images
+Virtual machines should not use preview versions of images.
 
 **Recommendation**: To use a non-preview image for virtual machines, in the [Microsoft.Compute/VirtualMachines/imageReference resource properties](https://docs.microsoft.com/en-us/azure/templates/microsoft.compute/virtualmachines?tabs=json#ImageReference), ensure the *version* property does not contain "-preview".
 
-## Ensure all virtual machine scale set instances are not using preview images
-Virtual Machine Scale Set instances should not use preview versions of images.
+## Ensure all Virtual Machine Scale Set instances are not using preview images
+Virtual machine scale set instances should not use preview versions of images.
```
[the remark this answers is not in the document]
yane3628: To make it more apparent in the titles which resources are being checked. Most of the ASC checks do this with resource names.

src/Analyzer.TemplateProcessor/ArmTemplateProcessor.cs
```diff
@@ -182,6 +173,133 @@ internal Template ProcessTemplateResourceAndOutputPropertiesLanguageExpressions(
...
+        {
+            Dictionary<string, TemplateResource > flattenedResources = new Dictionary<string, TemplateResource>(StringComparer.OrdinalIgnoreCase);
+
+            foreach (var resource in resources)
+            {
+                string dictionaryKey;
+
+                if (parentName != null && parentType != null)
+                {
+                    dictionaryKey = $"{parentName}/{resource.Name.Value} {parentType}/{resource.Type.Value}";
```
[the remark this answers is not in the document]
yane3628: Can you give an example?

src/Analyzer.Utilities/JsonLineNumberResolver.cs
yane3628: It's inaccessible at this scope, so I renamed it to match.

src/Analyzer.TemplateProcessor/ArmTemplateProcessor.cs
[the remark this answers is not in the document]
JohnathonMohr: Just discovered that the [documentation](https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/child-resource-name-type#within-parent-resource) states that child resources don't include their parent's type when specifying their own type.  (It doesn't say you *have* to leave that out, but I'll assume that's the case for now.)  So I think this is actually fine.

src/Analyzer.Utilities/JsonLineNumberResolver.cs
yane3628: Is there a reason you compare the lengths of the paths here instead of comparing the strings themselves? `originalTemplatePath.Length == pathInExpandedTemplate.Length`

JohnathonMohr: If the paths between original and expanded match, doing a length comparison is faster than a string comparison.  Because the resource (or param, variable, etc.) properties are going to have to match in both paths, it shouldn't be possible to have these two strings be the same length but have different paths.

In looking into this though, this doesn't correctly account for different lengths in integer indexes in the resource array though (if the original resource of a copied one has to be looked up).  Before your change, it would still happen to work because it wouldn't be possible to cut off the string exactly before a *resources* property.

So short conclusion: this logic could use some updating.

src/Analyzer.Utilities/JsonLineNumberResolver.cs
```diff
@@ -115,20 +117,108 @@ public int ResolveLineNumber(string pathInExpandedTemplate)
...
+                {
+                    var substringMatch = matches[i];
+
+                    if (i > 0)
+                    {
+                        resourceWithIndex += ".";
+                    }
+
+                    resourceWithIndex += substringMatch.Value;
+                }
```
JohnathonMohr, line 155: Suggestion (something like):
``` suggestion
                string resourceWithIndex = string.Join(".", matches);
```

src/Analyzer.Utilities/JsonLineNumberResolver.cs
```diff
@@ -115,20 +117,108 @@ public int ResolveLineNumber(string pathInExpandedTemplate)
...
+                    var childName = childResourceName.Value<string>();
+                    var childType = childResourceType.Value<string>();
+
+                    // Find the resource with this name and type in the expanded template
+                    string resourcePathInOriginalTemplate = GetResourcePath(originalTemplateRoot.InsensitiveToken("resources", InsensitivePathNotFoundBehavior.Null), childName, childType);
+                    
+                    return ResolveLineNumber($"{resourcePathInOriginalTemplate}.{remainingPathAtResourceScope}");
+                }
+            }
             return 1;
```
JohnathonMohr, line 180: Add a comment to make it clear this is an unexpected case

src/Analyzer.Utilities/JsonLineNumberResolver.cs
```diff
@@ -115,20 +117,108 @@ public int ResolveLineNumber(string pathInExpandedTemplate)
...
         }
+
+        /// <summary>
+        /// Given a list of resources, resource name, and resource type, return the JSON path of that resource.
+        /// </summary>
+        /// <param name="resources">JArray of resources.</param>
+        /// <param name="resourceName">Name of resource.</param>
+        /// <param name="resourceType">Type of resource.</param>
+        /// <returns>The JSON path where the resources is located.</returns>
```
JohnathonMohr, line 189: minor: extra 's' in 'resources'

src/Analyzer.Utilities/JsonLineNumberResolver.cs
```diff
@@ -115,20 +117,108 @@ public int ResolveLineNumber(string pathInExpandedTemplate)
...
+                var thisResourceName = resource.InsensitiveToken("name")?.Value<string>();
+                var thisResourceType = resource.InsensitiveToken("type")?.Value<string>();
+                if (string.Equals(resourceType, thisResourceType, StringComparison.OrdinalIgnoreCase) && string.Equals(resourceName, thisResourceName, StringComparison.OrdinalIgnoreCase))
+                {
+                    return resource.Path;
+                }
+
+                if (resource.InsensitiveToken("resources", InsensitivePathNotFoundBehavior.Null) != null)
+                {
+                    string path = GetResourcePath(resource.InsensitiveToken("resources"), resourceName, resourceType);
```
JohnathonMohr, line 208: minor: save resource.InsensitiveToken("resources") so it doesn't get fetched twice.

src/Analyzer.Utilities/JsonLineNumberResolver.cs
```diff
@@ -115,20 +117,108 @@ public int ResolveLineNumber(string pathInExpandedTemplate)
...
+                    
+                    if (path == null)
+                    {
+                        continue;
+                    }
+                    else
+                    {
+                        return path;
+                    }
```
JohnathonMohr, line 217: Suggestion:
``` suggestion
                    if (path != null)
                    {
                        return path;
                    }
```

src/Analyzer.Utilities/JsonLineNumberResolver.cs
```diff
@@ -115,20 +117,108 @@ public int ResolveLineNumber(string pathInExpandedTemplate)
...
 
+            bool pathsAreEqual = string.Equals(originalTemplatePath, pathInExpandedTemplate, StringComparison.OrdinalIgnoreCase);
+
+            bool unmatchedSegmentReferencesResourcesArray = unmatchedPathInOriginalTemplate.StartsWith("resources", StringComparison.OrdinalIgnoreCase);
+            bool unmatchedSegmentStartsWithIndexArray = unmatchedPathInOriginalTemplate.StartsWith("[", StringComparison.OrdinalIgnoreCase);
+            bool originalPathEndsWithResourcesArray = originalTemplatePath.EndsWith("resources", StringComparison.OrdinalIgnoreCase);
+
+            bool pathIsNotFromCopiedResource = !unmatchedSegmentReferencesResourcesArray
+                    && !unmatchedSegmentStartsWithIndexArray
+                    && !originalPathEndsWithResourcesArray;
```
JohnathonMohr, line 128: I don't think this is quite right.  For example, if the expanded path is *resources[0].dependsOn[3]* and the matched portion is *resources[0].dependsOn* (unmatched being *[3]*), `unmatchedSegmentStartsWithIndexArray` would be `true`, and it would continue and start looking for copied resources.  It should actually probably just return the line at *resources[0].dependsOn*.

src/Analyzer.Utilities/JsonLineNumberResolver.cs
```diff
@@ -115,20 +117,108 @@ public int ResolveLineNumber(string pathInExpandedTemplate)
             var originalTemplatePath = tokenFromOriginalTemplate.Path;
             var unmatchedPathInOriginalTemplate = pathInExpandedTemplate[originalTemplatePath.Length..].TrimStart('.');
```
JohnathonMohr, line 118: This needs to update too (same problem with different index string lengths).  You can simply create an issue for this if you don't want to fix it now.

JohnathonMohr (changes requested):

src/Analyzer.Utilities/JsonLineNumberResolver.cs
yane3628: In this example, where does dependsOn[3] come from if it's not in the original template?

JohnathonMohr: Maybe variable evaluation?  Might be unlikely for dependsOn, but could happen with any array.  For a vnet/subnet, maybe *resources[0].properties.subnets[0]*.

src/Analyzer.Utilities/JsonLineNumberResolver.cs
```diff
@@ -14,7 +14,8 @@ namespace Microsoft.Azure.Templates.Analyzer.Utilities
     /// </summary>
     public class JsonLineNumberResolver : ILineNumberResolver
     {
-        private static readonly Regex resourceIndexInPath = new Regex(@"resources\[(?<index>\d+)\]", RegexOptions.IgnoreCase | RegexOptions.Compiled);
+        private static string resourceIndexPattern = @"resources\[(?<index>\d+)\]";
+        private static readonly Regex resourceIndexInPath = new Regex(resourceIndexPattern, RegexOptions.IgnoreCase | RegexOptions.Compiled);
```
JohnathonMohr, line 18: This change can be reverted so the string is specified directly in the Regex creation.

src/Analyzer.Utilities/JsonLineNumberResolver.cs
```diff
@@ -115,20 +117,96 @@ public int ResolveLineNumber(string pathInExpandedTemplate)
...
+            if (pathsAreEqual || pathIsNotFromCopiedResource)
             {
                 return (tokenFromOriginalTemplate as IJsonLineInfo)?.LineNumber ?? 0;
             }
 
-            // The first unmatched property must be a resources[] array, likely a subresource
-            // of a top-level resource.  This is possible if resources were copied into other
-            // resources as part of template expansion.
-            // TODO
+            // Get the line number of the original resource before it was copied
```
JohnathonMohr, line 138: Any reason the original comment can't stay here too?  It helps explain why the algorithm continues.

JohnathonMohr (approved):

PR #128, opened 2021-06-14
Use full file path to read built-in rules file

### LINKED ISSUE #110: Rules are only loaded from current working directory
Running TemplateAnalyzer always looks for JSON rules to run from the current working directory - and no where else.  This means the analyzer must *always* be executed from its install location, instead of (ideally) where the target template or directory resides.

Instead, while the rule path currently can't be customized, the analyzer should *at least* look in the directory where the analyzer executable is, in addition to the working directory.

VeraBE:
Closes #110 

Now the executable can be run from any directory. We don't want to scan the working directory now for new rules as the issue suggests, further design is needed to support custom rule files

### DIFF
--- a/src/Analyzer.Core/TemplateAnalyzer.cs
+++ b/src/Analyzer.Core/TemplateAnalyzer.cs
@@ -3,6 +3,8 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
+using System.Reflection;
 using Microsoft.Azure.Templates.Analyzer.RuleEngines.JsonEngine;
 using Microsoft.Azure.Templates.Analyzer.TemplateProcessor;
 using Microsoft.Azure.Templates.Analyzer.Types;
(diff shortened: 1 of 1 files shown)


src/Analyzer.Core/TemplateAnalyzer.cs
```diff
@@ -75,7 +77,7 @@ public IEnumerable<IEvaluation> EvaluateRulesAgainstTemplate()
 
         private static string LoadRules()
         {
-            return System.IO.File.ReadAllText("Rules/BuiltInRules.json");
+            return System.IO.File.ReadAllText(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location) + "/Rules/BuiltInRules.json");
```
JohnathonMohr, line 80:
``` suggestion
            return File.ReadAllText(
                Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location) +
                "/Rules/BuiltInRules.json");
```

JohnathonMohr (approved):
Small suggestion, but looks good.

yane3628 (approved):

PR #129, opened 2021-06-14
New test project for validating built-in JSON rules

### LINKED ISSUE #40: Creating Test Infrastructure for Testing Built-in Rules
A new test project should be added that allows easy, config-based validations to be written for built-in rules.  This project will essentially provide a small test infrastructure for validating these rules.

At a high level, the project should:
- Run the template analyzer against templates specified in a config
- Validate the results of the run against the expectations specified in the config
- The config should specify (at least):
  - A template to run against
  - The rule that should flag a result (i.e. fail)
  - The expected line number of the failure

JohnathonMohr:
This adds a new test project for validating the correctness of built-in JSON rules.  Closes #40.
- Tests for rules are defined JSON.
- Test templates can be added to analyze for comparing with expected results.
- Test project does the work of discovering the test configurations, scanning the appropriate templates, and validating the rules.
- README file describes how tests are defined and run.


src/Analyzer.Core.JsonRuleTests/Analyzer.Core.JsonRuleTests.csproj
```diff
@@ -0,0 +1,29 @@
...
+    <PackageReference Include="MSTest.TestFramework" Version="2.2.3" />
+    <PackageReference Include="coverlet.collector" Version="3.0.2" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Analyzer.Core\Analyzer.Core.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <None Update="Tests\*">
```
VeraBE, line 21: Maybe we want a folder per built-in rule, to store the test templates and the file with the test information? Because we would like minimal test templates, right? So there shouldn't be that much template reuse

src/Analyzer.Core.JsonRuleTests/README.md
```diff
@@ -0,0 +1,46 @@
...
+        "Template": "Name of template file to analyze (without file extension).",
+        "ExpectPass": "true/false - Whether the template named in 'Template' is expected to pass for this rule.",
+        "ReportedLines": "Array of integers - If 'ExpectPass' is false, these are the line numbers expected to be reported in the failure."
+    },
+    ... // More tests can be defined if multiple templates should be analyzed - one test block for each template
+]
+```
+
+### Test ARM Templates
+For each template file referenced in a `Template`, there should be a file in the *TestTemplates* directory with the same name, having a ".badtemplate" file extension.  This extension is used to help prevent these templates from actually being deployed in Azure.
```
VeraBE, line 39: We want to trade all the comforts of writing and reading a JSON file with proper syntax highlighting just for the possibility of someone thinking a minimal test file is a production-ready template?

src/Analyzer.Core.JsonRuleTests/README.md
```diff
@@ -0,0 +1,46 @@
+# Test Project for Built-in JSON Rules
+This project is designed for quickly adding tests to verify the correctness of built-in JSON rules.  When new rules are added, tests should be added here to make sure the rule is written correctly and help protect against incorrect changes to the rule.
+
+## Test Setup
+A test (or tests) for a given JSON rule consists of 2 parts:
+1. A JSON configuration describing the test(s).
+2. One or more test ARM templates to analyze.  The results of the analysis are compared with the test configuration to assert correctness of the JSON rule.
+
+### JSON Configuration
+To create a set of tests for a JSON rule, a new JSON file is created in the *Tests* directory.  The name of the file must be the same as the `name` property of the JSON rule, with ".json" as the file extension.
```
VeraBE, line 10: I was thinking if it makes sense to add the logic to have a unique id for each rule as part of this work? Because I suppose we will start creating a ton of new rules using this framework and it might be harder to add after that. We could also use the ID for these file names

src/Analyzer.Core.JsonRuleTests/README.md
```diff
@@ -0,0 +1,46 @@
+# Test Project for Built-in JSON Rules
```
VeraBE, line 1: Wouldn't this also work for PowerShell rules? Seems like a generic framework that could potentially work with any rule type inside of the tool

src/Analyzer.Core.JsonRuleTests/Tests/AppServiceApiApp_HTTPS.json
```diff
@@ -0,0 +1,7 @@
+[
```
yane3628, line 1: It would be nice if we can add some info on what scenario the result would fall under. For example: 
- The line X result is in the case where it's a finding in a dependsOn resource. 
- Line Y is a result from the base case of the evaluation. 
- Line Z is the case where 1 of 3 leaf expressions failed in an allOf operator.

src/Analyzer.Core.JsonRuleTests/Tests/AppServiceApiApp_HTTPS.json
```diff
@@ -0,0 +1,7 @@
...
+    {
```
yane3628, line 2: I think we should be able to add tests for ExpectedPass = true

src/Analyzer.Core.JsonRuleTests/Analyzer.Core.JsonRuleTests.csproj
```diff
@@ -0,0 +1,29 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
yane3628, line 1: Add the xml comment errors to the project

src/Analyzer.Core.JsonRuleTests/README.md
yane3628, line 10: "The name of the file must be the same as the `name` property of the JSON rule" should probably be bold

src/Analyzer.Core.JsonRuleTests/README.md
```diff
@@ -0,0 +1,46 @@
...
+1. A JSON configuration describing the test(s).
+2. One or more test ARM templates to analyze.  The results of the analysis are compared with the test configuration to assert correctness of the JSON rule.
+
+### JSON Configuration
+To create a set of tests for a JSON rule, a new JSON file is created in the *Tests* directory.  The name of the file must be the same as the `name` property of the JSON rule, with ".json" as the file extension.
+
+For example, a JSON rule like the following:
+``` json
+{
+    "name": "SuperSecuryCheck",
```
yane3628, line 15: *Security also in the name of the file after the code snippet.

src/Analyzer.Core.JsonRuleTests/README.md
```diff
@@ -0,0 +1,46 @@
+# Test Project for Built-in JSON Rules
+This project is designed for quickly adding tests to verify the correctness of built-in JSON rules.  When new rules are added, tests should be added here to make sure the rule is written correctly and help protect against incorrect changes to the rule.
+
+## Test Setup
+A test (or tests) for a given JSON rule consists of 2 parts:
```
yane3628, line 5: I feel like this description can be reworded to be clearer. When I read this intro section, I didn't follow without seeing an example of a test file at the same time and didn't realize that the second point was the templates in the TestTemplates folder until I got the end of this README...

src/Analyzer.Core.JsonRuleTests/README.md
```diff
@@ -0,0 +1,46 @@
...
+}
+```
+... would have tests defined in a configuration file named "Tests/SuperSecuryCheck.json".
+
+The JSON test configuration has the following schema:
+``` json
+[
+    {
+        "Template": "Name of template file to analyze (without file extension).",
+        "ExpectPass": "true/false - Whether the template named in 'Template' is expected to pass for this rule.",
```
yane3628, line 31: You might want to add descriptions as comments instead of the string value. This would make it more clear that ExpectPass is a boolean type.

src/Analyzer.Core.JsonRuleTests/README.md
yane3628: Is there a reason the template field doesn't include the file extension? I feel like it would be easier to follow if it did.

src/Analyzer.Core.JsonRuleTests/TestRunner.cs
```diff
@@ -0,0 +1,115 @@
...
+using Microsoft.Azure.Templates.Analyzer.Types;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.Templates.Analyzer.Core.JsonRuleTests
+{
+    [TestClass]
+    public class TestRunner
+    {
+        Dictionary<string, IEnumerable<IEvaluation>> templateEvaluations = new();
```
yane3628, line 18: Can we use the full constructor syntax? I personally think it's a little confusing understanding `new()` Thoughts @VeraBE

src/Analyzer.Core.JsonRuleTests/TestRunner.cs
```diff
@@ -0,0 +1,115 @@
...
+                {
+                    // Get all lines reported as failed
+                    var failingLines = GetAllResults(evaluation)
+                        .Where(r => !r.Passed)
+                        .Select(r => r.LineNumber).ToHashSet();
+
+                    // Verify all expected lines are reported
+                    var expectedLines = ruleExpectations.ReportedLines.ToHashSet();
+                    Assert.IsTrue(failingLines.SetEquals(expectedLines),
+                        "Expected failing lines do not match actual failed lines.  " +
```
yane3628, line 63: I did not know you can add a message to Assert.IsTrue. Cool! This is useful for all the tests we do DeepEquals on JSON objects.

src/Analyzer.Core.JsonRuleTests/TestRunner.cs
VeraBE: I like `var templateEvaluations = new  Dictionary<string, IEnumerable<IEvaluation>>()` better too

VeraBE:
```suggestion
        var templateEvaluations = new Dictionary<string, IEnumerable<IEvaluation>>()
```

VeraBE: (I wanted to test this new suggestion feature 😂)

src/Analyzer.Core.JsonRuleTests/Tests/AppServiceApiApp_HTTPS.json
JohnathonMohr: I did sort of add some examples, but they are implicit.  There are resources defined in the test template that pass, and there are no line numbers reported within those resources.  ExpectPass could be true if a template is added for that purpose.  I didn't do that here.

src/Analyzer.Core.JsonRuleTests/Analyzer.Core.JsonRuleTests.csproj
JohnathonMohr: Those are only supposed to be for product code (not test code)

src/Analyzer.Core.JsonRuleTests/README.md
JohnathonMohr: Not including the extension in the Template field allows enforcement of the '.badtemplate' extension.  Whether we keep that requirement is still up for discussion though.

src/Analyzer.Core.JsonRuleTests/README.md
JohnathonMohr: I considered that.  But the JSON parsing will parse a bool from a string without issues too, so I figured it didn't really matter much.

src/Analyzer.Core.JsonRuleTests/TestRunner.cs
JohnathonMohr: Pretty much all the functions in the `Assert` class have an optional message argument :)

src/Analyzer.Core.JsonRuleTests/TestRunner.cs
JohnathonMohr: Two things:
- the `new()` syntax has been used in other places too.  Do we need to agree on a convention when it's ok to use?
  - Personally, I think it will be easy to get used to.  It might be confusing the first time seeing it, but you only need to gain understanding once.
- `var` can't be used for class members - either the full `Dictionary<...>` type needs to be written out twice, or we use `new()`.

src/Analyzer.Core.JsonRuleTests/Tests/AppServiceApiApp_HTTPS.json
JohnathonMohr: Do you mean adding comments like that in the test config?

src/Analyzer.Core.JsonRuleTests/README.md
JohnathonMohr: Good call.

src/Analyzer.Core.JsonRuleTests/README.md
JohnathonMohr: Yes, relying on unique ids here would be great.  That needs to be done soon, but I don't think we should combine that with this change.  We still need to agree on what will look like.

src/Analyzer.Core.JsonRuleTests/README.md
```diff
@@ -0,0 +1,46 @@
...
+        "ReportedLines": "Array of integers - If 'ExpectPass' is false, these are the line numbers expected to be reported in the failure."
```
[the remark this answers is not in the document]
JohnathonMohr: Do you mean validating the messages that are returned in a result?

src/Analyzer.Core.JsonRuleTests/README.md
JohnathonMohr: I don't know if it's needed or not.  This seemed like a good safety measure.  These files can always be renamed to .json while editing, then renamed back.

src/Analyzer.Core.JsonRuleTests/Analyzer.Core.JsonRuleTests.csproj
JohnathonMohr: I organized them this way because I wanted to avoid a bunch of nested directories - it looks cleaner to me being just one directory.

The effort involved in creating test templates isn't something I have a clear picture on yet.  Ideally as little effort as possible.  Does that mean many simple templates?  Or fewer templates that define everything tests need?  Probably a combination of both.  I don't want to prevent tests from taking advantage of test templates that already exist that provide everything needed.

Template naming also needs to be considered.  When multiple templates contain almost identical contents, but differ slightly for test purposes, how can those be differentiated in a meaningful way without having almost sentence-long names?  This was one reason I considered the desire to combine some scenarios into single templates.

src/Analyzer.Core.JsonRuleTests/Tests/AppServiceApiApp_HTTPS.json
yane3628: Yea comments would work I think.

src/Analyzer.Core.JsonRuleTests/Tests/AppServiceApiApp_HTTPS.json
yane3628: I know the TTK has templates with expected fail scenarios and expected pass scenarios in different folders.

src/Analyzer.Core.JsonRuleTests/Analyzer.Core.JsonRuleTests.csproj
VeraBE: mmm I can see how having test templates ready could simplify writing some test cases, but for edge cases I imagine we would end up creating specific long name templates anyways

src/Analyzer.Core.JsonRuleTests/README.md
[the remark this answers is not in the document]
VeraBE: I thought at some point we were going to design a way to handle extra information, so if the tool is consumed as a library the user doesn't have to parse the error message, and it's easier to improve how we show it in the CLI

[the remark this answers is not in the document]
VeraBE: I was thinking we might want to do something like:
```
{
	"Template": "Name of template file to analyze (without file extension).",
	"ExpectPass": "true/false - Whether the template named in 'Template' is expected to pass for this rule.",
	"ReportedErrors": [
		{
			"LineNumber": 5
		}
	]
}
```
Or something like that, in case later we want to do something like:

```
{
	"Template": "Name of template file to analyze (without file extension).",
	"ExpectPass": "true/false - Whether the template named in 'Template' is expected to pass for this rule.",
	"ReportedErrors": [
		{
			"LineNumber": 5,
			"VariableName": ""
		},
		{
			"FlaggedString": ""
		}
	]
}
```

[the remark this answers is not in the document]
VeraBE: Writing that example made me think that specifying ExpectPass and an array of lines/errors is kind of unnecessary, an empty list could be interpreted as a test with no errors

src/Analyzer.Core.JsonRuleTests/TestRunner.cs
VeraBE: I didn't notice it was a class member! Personally, I don't think we need a convention for that, but I like using var every time it's possible, to focus on the variable name

src/Analyzer.Core.JsonRuleTests/README.md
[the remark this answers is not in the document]
JohnathonMohr: I like your thinking about this.  At least changing to an array of objects with a LineNumber makes sense.  Then we can add more once we decide how to surface the information.

src/Analyzer.Core.JsonRuleTests/Tests/AppServiceApiApp_HTTPS.json
yane3628: @JohnathonMohr Can you add comments for each expected result for the tests you've added?

yane3628 (approved):

VeraBE (approved):

PR #130, opened 2021-06-15
Fix copy resources line number mapping bug

### LINKED ISSUE #125: ResolveLineNumber in the copy resources scenario does not account for the case where the expanded template resource name is not the same as the original template resource name
The expanded resource name can be different from the the original template resource name because the original resource name can refer to a parameter or variable, etc.

yane3628:
- Makes and uses mapping between resources in expanded template to original template for easier line number mapping (Closes #125)
- Processes all resources language expressions once flattened

### DIFF
--- a/src/Analyzer.Core/TemplateAnalyzer.cs
+++ b/src/Analyzer.Core/TemplateAnalyzer.cs
@@ -47,7 +47,7 @@ namespace Microsoft.Azure.Templates.Analyzer.Core
             var armTemplateProcessor = new ArmTemplateProcessor(Template);
 
             try
-            {   
+            {
                 templatejObject = armTemplateProcessor.ProcessTemplate(Parameters);
             }
             catch (Exception e)
--- a/src/Analyzer.Core/TemplateAnalyzer.cs
+++ b/src/Analyzer.Core/TemplateAnalyzer.cs
@@ -44,11 +44,10 @@ namespace Microsoft.Azure.Templates.Analyzer.Core
         public IEnumerable<IEvaluation> EvaluateRulesAgainstTemplate()
         {
             JToken templatejObject;
-            ArmTemplateProcessor armTemplateProcessor;
+            var armTemplateProcessor = new ArmTemplateProcessor(Template);
 
             try
-            {
-                armTemplateProcessor = new ArmTemplateProcessor(Template);
+            {   
                 templatejObject = armTemplateProcessor.ProcessTemplate(Parameters);
             }
             catch (Exception e)
--- a/src/Analyzer.Core/TemplateAnalyzer.cs
+++ b/src/Analyzer.Core/TemplateAnalyzer.cs
@@ -37,10 +37,11 @@ namespace Microsoft.Azure.Templates.Analyzer.Core
         public IEnumerable<IEvaluation> EvaluateRulesAgainstTemplate()
         {
             JToken templatejObject;
+            ArmTemplateProcessor armTemplateProcessor;
 
             try
             {
-                ArmTemplateProcessor armTemplateProcessor = new ArmTemplateProcessor(Template);
+                armTemplateProcessor = new ArmTemplateProcessor(Template);
                 templatejObject = armTemplateProcessor.ProcessTemplate(Parameters);
             }
             catch (Exception e)
@@ -62,7 +63,8 @@ namespace Microsoft.Azure.Templates.Analyzer.Core
                     new TemplateContext {
                         OriginalTemplate = JObject.Parse(Template),
                         ExpandedTemplate = templatejObject,
-                        IsMainTemplate = true },
+                        IsMainTemplate = true,
+                        ResourceMappings = armTemplateProcessor.ResourceMappings },
                     rules);
 
                 return evaluations;
--- a/src/Analyzer.TemplateProcessor.UnitTests/ArmTemplateProcessorTests.cs
+++ b/src/Analyzer.TemplateProcessor.UnitTests/ArmTemplateProcessorTests.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Collections.Generic;
 using Azure.Deployments.Core.Collections;
 using Azure.Deployments.Core.Extensions;
 using Azure.Deployments.Core.Json;
@@ -18,8 +19,8 @@ namespace Microsoft.Azure.Templates.Analyzer.TemplateProcessor.UnitTests
     {
         private static ArmTemplateProcessor _armTemplateProcessor;
 
-        [ClassInitialize]
-        public static void ClassInit(TestContext testContext)
+        [TestInitialize]
+        public void TestInit()
         {
             _armTemplateProcessor = new ArmTemplateProcessor(@"{
                 ""$schema"": ""https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#"",
@@ -181,10 +182,16 @@ namespace Microsoft.Azure.Templates.Analyzer.TemplateProcessor.UnitTests
                 ""outputs"": { }
             }";
 
+            Dictionary<string, string> expectedMapping = new Dictionary<string, string> {
+                { "resources[0]", "resources[0]" }
+            };
+
             Template template = TemplateEngine.ParseTemplate(templateJson);
             _armTemplateProcessor.ProcessResourcesAndOutputs(template);
 
             Assert.IsTrue(template.Resources.First().Properties.Value.InsensitiveToken("dnsSettings.domainNameLabel").Value<string>().StartsWith("linux-vm-"));
+
+            AssertDictionariesAreEqual(expectedMapping, _armTemplateProcessor.ResourceMappings);
         }
 
         [DataTestMethod]
@@ -194,10 +201,14 @@ namespace Microsoft.Azure.Templates.Analyzer.TemplateProcessor.UnitTests
         {
             string templateJson = GenerateTemplate(outputValue);
 
+            Dictionary<string, string> expectedMapping = new Dictionary<string, string>();
+
             Template template = TemplateEngine.ParseTemplate(templateJson);
             _armTemplateProcessor.ProcessResourcesAndOutputs(template);
 
             Assert.IsTrue(template.Outputs["output0"].Value.Value.ToString().StartsWith(expectedValue));
+
+            AssertDictionariesAreEqual(expectedMapping, _armTemplateProcessor.ResourceMappings);
         }
 
         [TestMethod]
@@ -218,9 +229,15 @@ namespace Microsoft.Azure.Templates.Analyzer.TemplateProcessor.UnitTests
                 { "tenantId", "00000000-0000-0000-0000-000000000000" }
             };
 
+            Dictionary<string, string> expectedMapping = new Dictionary<string, string> {
+                { "resources[0]", "resources[0]" }
+            };
+
             Template template = _armTemplateProcessor.ParseAndValidateTemplate(parameters, metadata);
 
             Assert.IsTrue(template.Resources.First().Properties.Value.InsensitiveToken("dnsSettings.domainNameLabel").Value<string>().StartsWith("linux-vm-"));
+
+            AssertDictionariesAreEqual(expectedMapping, _armTemplateProcessor.ResourceMappings);
         }
 
         [TestMethod]
@@ -272,6 +289,11 @@ namespace Microsoft.Azure.Templates.Analyzer.TemplateProcessor.UnitTests
                 { "tenantId", "00000000-0000-0000-0000-000000000000" }
             };
 
+            Dictionary<string, string> expectedMapping = new Dictionary<string, string> {
+                { "resources[0]", "resources[0]" },
+                { "resources[1]", "resources[0]" }
+            };
+
             Template template = armTemplateProcessor.ParseAndValidateTemplate(parameters, metadata);
 
             Assert.AreEqual(2, template.Resources.Length);
@@ -283,6 +305,8 @@ namespace Microsoft.Azure.Templates.Analyzer.TemplateProcessor.UnitTests
             Assert.AreEqual("name/SSH-VMNamePrefix-1", template.Resources.Last().Name.Value);
             Assert.AreEqual("[concat('name', '/', 'SSH-', 'VMNamePrefix-', copyIndex())]", template.Resources.Last().OriginalName);
             Assert.AreEqual(2201, template.Resources.Last().Properties.Value.InsensitiveToken("frontendPort").Value<int>());
+
+            AssertDictionariesAreEqual(expectedMapping, armTemplateProcessor.ResourceMappings);
         }
 
         [TestMethod]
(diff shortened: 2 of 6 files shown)


src/Analyzer.TemplateProcessor/ArmTemplateProcessor.cs
```diff
@@ -219,58 +235,98 @@ internal void CopyResourceDependants(TemplateResource templateResource, Dictiona
...
+        private void AddResourceMapping(string expandedTemplatePath, string originalTemplatePath)
+        {
+            // Save all permutations of the resource path based off values already present 
+            // in the dictionary with mapping. This is necessary to report an issue in
+            // a copied nth grandchild resource.
+            var tokens = expandedTemplatePath.Split('.');
+
+            for (int i = 0; i < tokens.Length - 1; i++)
+            {
+                string subPath = string.Join('.', tokens[..(i + 1)]);
```
JohnathonMohr, line 263: Suggested name change: `potentialOriginalPathOfAnotherResource`?

src/Analyzer.TemplateProcessor/ArmTemplateProcessor.cs
```diff
@@ -219,58 +235,98 @@ internal void CopyResourceDependants(TemplateResource templateResource, Dictiona
...
+                if (originalToExpandedMapping.ContainsKey(subPath))
+                {
+                    foreach (string path in originalToExpandedMapping[subPath])
```
JohnathonMohr, line 266: Suggested name change: `copiedLocationOfAnotherResource`?

src/Analyzer.TemplateProcessor/ArmTemplateProcessor.cs
```diff
@@ -219,58 +235,98 @@ internal void CopyResourceDependants(TemplateResource templateResource, Dictiona
...
+                    {
```
JohnathonMohr, line 267: (With name changes):
``` suggestion
                if (originalToExpandedMapping.TryGetValue(potentialOriginalPathOfAnotherResource, out copiedLocationsOfAnotherResource))
                {
                    foreach (var copiedLocationOfAnotherResource in copiedLocationsOfAnotherResource)
                    {
```

src/Analyzer.TemplateProcessor/ArmTemplateProcessor.cs
```diff
@@ -378,6 +426,12 @@ private TemplateResource[] ReorderResourceCopies(Template template, Dictionary<s
                         else if (!dropResourceCopies)
                         {
                             updatedOrder.Add(resource);
+                            resource.Path = $"resources[{updatedOrder.Count - 1}]";
```
JohnathonMohr, line 429: I'm realizing now this will break if there are 2 or more resources with their own copy loops.  Both will be added to the end by the parsing and expansion, then reordering will add mappings for the copies of the first resource (below), then the second resource with a copy loop will be inserted into the array above (with `CopyIndex == 0`), pushing out the previous copies an extra index, making their previous mappings invalid.

JohnathonMohr (changes requested):
Most are minor changes, but there's one bug that needs to be solved.

VeraBE (approved):

src/Analyzer.TemplateProcessor.UnitTests/ArmTemplateProcessorTests.cs
```diff
@@ -240,6 +240,115 @@ public void ParseAndValidateTemplate_ValidTemplateWithExpressionInResourceProper
...
+            // First resource should be the resource without a copy index specified
+            Assert.AreEqual("resourceWith/NoCopyLoop", template.Resources.First().Name.Value);
+
+            Assert.AreEqual("name/SSH-VMNamePrefix-1", template.Resources[2].Name.Value);
+            Assert.AreEqual("[concat('name', '/', 'SSH-', 'VMNamePrefix-', copyIndex())]", template.Resources[2].OriginalName);
+            Assert.AreEqual(2201, template.Resources[2].Properties.Value.InsensitiveToken("frontendPort").Value<int>());
+
+            Assert.AreEqual("name/CopyLoop2-SSH-VMNamePrefix-1", template.Resources.Last().Name.Value);
+            Assert.AreEqual("[concat('name', '/', 'CopyLoop2-SSH-', 'VMNamePrefix-', copyIndex())]", template.Resources.Last().OriginalName);
+            Assert.AreEqual(2201, template.Resources.Last().Properties.Value.InsensitiveToken("frontendPort").Value<int>());
```
JohnathonMohr, line 347: I don't think these assertions provide much value.  Since we have the path mapping between expanded and original templates (and that's all that's used), the order of the resources in the expanded template shouldn't matter in the end.

src/Analyzer.TemplateProcessor.UnitTests/ArmTemplateProcessorTests.cs
```diff
@@ -418,13 +527,6 @@ public void CopyResourceDependants_ValidChildResourceDependsOnByNameAndResourceI
                 ""dependsOn"": [ ""someResource"" ],
                 ""properties"": { }
             }", DisplayName = "Child resource depends on resource outside template scope")]
-        [DataRow(@"{
-                ""type"": ""Microsoft.Network/networkInterfaces"",
-                ""apiVersion"": ""2018-10-01"",
-                ""name"": ""simpleLinuxVMNetInt"",
-                ""location"": ""westus"",
-                ""properties"": { }
-            }", DisplayName = "Child resource depends on not specified")]
```
JohnathonMohr, line 380: Why was this test removed?

src/Analyzer.TemplateProcessor/ArmTemplateProcessor.cs
```diff
@@ -392,58 +391,47 @@ private TemplateExpressionEvaluationHelper GetTemplateFunctionEvaluationHelper(T
...
-                            updatedOrder.Insert(originalValues.Item2, resource);
-                        }
-                        else if (!dropResourceCopies)
-                        {
-                            updatedOrder.Add(resource);
-                            resource.Path = $"resources[{updatedOrder.Count - 1}]";
-                        }
 
-                        if (!dropResourceCopies)
+                        if (!dropResourceCopies || resource.CopyContext.CopyIndex == 0)
```
JohnathonMohr, line 416: I think we should remove the drop-copy logic.  I don't envision it being used.

src/Analyzer.TemplateProcessor/ArmTemplateProcessor.cs
```diff
@@ -392,58 +391,47 @@ private TemplateExpressionEvaluationHelper GetTemplateFunctionEvaluationHelper(T
...
-                    updatedOrder.Add(resource);
-                }
-                else
-                {
-                    // Copied resource.  Update OriginalName and:
-                    // - if it's the first copy, insert back where it was supposed to be
-                    // - if it's an extra copy, add to the end, or don't add at all if requested
+                    // Copied resource.  Update OriginalName and
+                    // add maping to original resource
                     if (copyNameMap.TryGetValue(resource.Copy.Name.Value, out (string, int) originalValues))
```
JohnathonMohr, line 412: This should have paired 'else' logic

src/Analyzer.TemplateProcessor/ArmTemplateProcessor.cs
```diff
@@ -392,58 +391,47 @@ private TemplateExpressionEvaluationHelper GetTemplateFunctionEvaluationHelper(T
...
                             AddResourceMapping(resource.Path, $"resources[{originalValues.Item2}]");
                         }
-                    }
-                    else
-                    {
-                        // Couldn't get original values.  Insert at end as a precaution,
-                        // but this code should never be reached under normal circumstances.
-                        updatedOrder.Add(resource);
+
+                        continue;
```
JohnathonMohr, line 423: It looks like this only skips one line of code.  That line should be moved into the 'else' block so `continue` can be removed.

src/Analyzer.TemplateProcessor/ArmTemplateProcessor.cs
JohnathonMohr: Then this function doesn't have to return anything, it just maps.

src/Analyzer.TemplateProcessor/ArmTemplateProcessor.cs
```diff
@@ -392,58 +391,47 @@ private TemplateExpressionEvaluationHelper GetTemplateFunctionEvaluationHelper(T
...
-                    // non-copied resource.  Add to updated array
-                    updatedOrder.Add(resource);
-                }
-                else
-                {
-                    // Copied resource.  Update OriginalName and:
-                    // - if it's the first copy, insert back where it was supposed to be
-                    // - if it's an extra copy, add to the end, or don't add at all if requested
+                    // Copied resource.  Update OriginalName and
+                    // add maping to original resource
```
JohnathonMohr, line 411: minor: mapping

src/Analyzer.TemplateProcessor/ArmTemplateProcessor.cs
```diff
@@ -131,7 +131,7 @@ internal Template ParseAndValidateTemplate(InsensitiveDictionary<JToken> paramet
                 // Do not throw if there was an issue with evaluating language expressions
             }
 
-            template.Resources = ReorderResourceCopies(template, copyNameMap);
+            template.Resources = MapResources(template, copyNameMap);
```
JohnathonMohr, line 134: Since flatted (sub-) resources also have to be mapped later (line 164), should this method be named `MapTopLevelResources`?

src/Analyzer.TemplateProcessor.UnitTests/ArmTemplateProcessorTests.cs
```diff
@@ -479,12 +582,14 @@ public void CopyResourceDependants_DependsOnIsSubResource_CopiesToCorrectResourc
...
                   ""properties"": { },
+                  ""dependsOn"": [ ],
                   ""resources"": [
                     {
                       ""apiVersion"": ""2015-08-01"",
                       ""name"": ""web"",
                       ""type"": ""sourcecontrols"",
-                      ""properties"": { }
+                      ""properties"": { },
+                      ""dependsOn"": [ ],
```
JohnathonMohr, line 592: Why is an empty `"dependsOn"` array needed in so many resources?

src/Analyzer.Core/TemplateAnalyzer.cs
```diff
@@ -44,11 +44,10 @@ public TemplateAnalyzer(string template, string parameters = null, string templa
         public IEnumerable<IEvaluation> EvaluateRulesAgainstTemplate()
         {
             JToken templatejObject;
-            ArmTemplateProcessor armTemplateProcessor;
+            var armTemplateProcessor = new ArmTemplateProcessor(Template);
 
             try
-            {
-                armTemplateProcessor = new ArmTemplateProcessor(Template);
+            {
```
JohnathonMohr, line 50: minor: extra spaces after `{`

JohnathonMohr (changes requested):

src/Analyzer.TemplateProcessor.UnitTests/ArmTemplateProcessorTests.cs
yane3628: I can remove them and just leave the dictionary comparison.

src/Analyzer.TemplateProcessor.UnitTests/ArmTemplateProcessorTests.cs
yane3628: I realized it wasn't necessary. I had to update the test to call a function higher in the stack. Once this happened, the resource objects all had an empty dependsOn added to them after processing.

JohnathonMohr: I guess I won't block on it, but I think the test still has value - it verifies this scenario is covered in the parsing.  The implementation depends on the ARM Deployment library setting a default empty array, but that's a detail that's abstracted from the test.  i.e. the test doesn't care *how* it's accounted for, just that it is in some way.

src/Analyzer.TemplateProcessor.UnitTests/ArmTemplateProcessorTests.cs
yane3628: See my response in your other comment "Why was this test removed?"

src/Analyzer.TemplateProcessor.UnitTests/ArmTemplateProcessorTests.cs
yane3628: It seemed like unnecessary work to add the expectation of dependsOn being added when comparing the json at the end of the test. That's why I removed the test instead of updating it.

src/Analyzer.TemplateProcessor.UnitTests/ArmTemplateProcessorTests.cs
JohnathonMohr: It sounds like this property isn't actually necessary in the JSON then.  I guess I'm confused why it was added everywhere if it doesn't need to be specified.

src/Analyzer.TemplateProcessor.UnitTests/ArmTemplateProcessorTests.cs
JohnathonMohr: Oh, now I see what's going on.

I get the desire to simplify the tests.  I think we should when we can.  In this case though, doesn't it seem like it's removing a test scenario in favor of simplifying the test?  Now it looks like the parsing library requires "dependsOn" to be specified.

src/Analyzer.TemplateProcessor/ArmTemplateProcessor.cs
```diff
@@ -344,52 +388,36 @@ private TemplateExpressionEvaluationHelper GetTemplateFunctionEvaluationHelper(T
...
-                    updatedOrder.Add(resource);
-                }
-                else
-                {
-                    // Copied resource.  Update OriginalName and:
-                    // - if it's the first copy, insert back where it was supposed to be
-                    // - if it's an extra copy, add to the end, or don't add at all if requested
+                    // Copied resource.  Update OriginalName and
+                    // add mapping to original resource
                     if (copyNameMap.TryGetValue(resource.Copy.Name.Value, out (string, int) originalValues))
```
JohnathonMohr, line 406: Combine this with the outer `if`

src/Analyzer.TemplateProcessor/ArmTemplateProcessor.cs
```diff
@@ -344,52 +388,36 @@ private TemplateExpressionEvaluationHelper GetTemplateFunctionEvaluationHelper(T
...
+
+                        continue;
                     }
                 }
+                    
+                AddResourceMapping($"resources[{i}]", resource.Path);
             }
             
-            return updatedOrder.ToArray();
+            return;
```
JohnathonMohr, line 420: Unnecessary

src/Analyzer.TemplateProcessor.UnitTests/ArmTemplateProcessorTests.cs
yane3628: Added the test back and updated the test

JohnathonMohr (approved):

PR #131, opened 2021-06-15
When no scopes were found in structured expressions, do not create an evaluation

### LINKED ISSUE #124: Structured expressions like anyOf and allOf should ignore any leaf expressions that don't have results because no matching scopes were found
For example

```
"resoourceType": "Microsoft.Web/sites",
"anyOf": [
  {
    "resourceType": "Microsoft.Web/sites/config",
    "path": "some.path",
    "equals": "somevalue"
  },
  {
    "path": "someOther.path",
    "equals": "somevalue"
  }
]
```
The above should evaluate to false since `"Microsoft.Web/sites/config"` is not specified and `"someOther.path"` is not equal to `somevalue` in the below resource example.

```
{
  "type": "Microsoft.Web/sites",
  "someOther": {
    "property": "someOtherValue"
  },
  "resources": [ ]
}
```

yane3628:
When no scopes were found in structured expressions, do not create an evaluation (Closes #124)

### DIFF
--- a/src/Analyzer.JsonRuleEngine.UnitTests/AnyOfExpressionTests.cs
+++ b/src/Analyzer.JsonRuleEngine.UnitTests/AnyOfExpressionTests.cs
@@ -10,29 +10,13 @@ using Microsoft.Azure.Templates.Analyzer.Types;
 using Microsoft.Azure.Templates.Analyzer.Utilities;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
+using static Microsoft.Azure.Templates.Analyzer.RuleEngines.JsonEngine.UnitTests.TestUtilities;
 
 namespace Microsoft.Azure.Templates.Analyzer.RuleEngines.JsonEngine.UnitTests
 {
     [TestClass]
     public class AnyOfExpressionTests
     {
-        /// <summary>
-        /// A mock implementation of an <see cref="Expression"/> for testing internal methods.
-        /// </summary>
-        private class MockExpression : Expression
-        {
-            public Func<IJsonPathResolver, JsonRuleEvaluation> EvaluationCallback { get; set; }
-
-            public MockExpression(ExpressionCommonProperties commonProperties)
-                : base(commonProperties)
-            { }
-
-            public override JsonRuleEvaluation Evaluate(IJsonPathResolver jsonScope)
-            {
-                return base.EvaluateInternal(jsonScope, EvaluationCallback);
-            }
-        }
-
         [DataTestMethod]
         [DataRow(true, true, DisplayName = "AnyOf evaluates to true (true || true)")]
         [DataRow(true, false, DisplayName = "AnyOf evaluates to true (true || false)")]
--- a/src/Analyzer.JsonRuleEngine.UnitTests/AnyOfExpressionTests.cs
+++ b/src/Analyzer.JsonRuleEngine.UnitTests/AnyOfExpressionTests.cs
@@ -16,6 +16,23 @@ namespace Microsoft.Azure.Templates.Analyzer.RuleEngines.JsonEngine.UnitTests
     [TestClass]
     public class AnyOfExpressionTests
     {
+        /// <summary>
+        /// A mock implementation of an <see cref="Expression"/> for testing internal methods.
+        /// </summary>
+        private class MockExpression : Expression
+        {
+            public Func<IJsonPathResolver, JsonRuleEvaluation> EvaluationCallback { get; set; }
+
+            public MockExpression(ExpressionCommonProperties commonProperties)
+                : base(commonProperties)
+            { }
+
+            public override JsonRuleEvaluation Evaluate(IJsonPathResolver jsonScope)
+            {
+                return base.EvaluateInternal(jsonScope, EvaluationCallback);
+            }
+        }
+
         [DataTestMethod]
         [DataRow(true, true, DisplayName = "AnyOf evaluates to true (true || true)")]
         [DataRow(true, false, DisplayName = "AnyOf evaluates to true (true || false)")]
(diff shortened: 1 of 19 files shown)


src/Analyzer.JsonRuleEngine.UnitTests/AnyOfExpressionTests.cs
```diff
@@ -16,6 +16,23 @@ namespace Microsoft.Azure.Templates.Analyzer.RuleEngines.JsonEngine.UnitTests
     [TestClass]
     public class AnyOfExpressionTests
     {
+        /// <summary>
+        /// A mock implementation of an <see cref="Expression"/> for testing internal methods.
+        /// </summary>
+        private class MockExpression : Expression
```
VeraBE, line 22: Maybe is worth moving this to its own file so there's no repeated code between AnyOfExpressionTests and AllOfExpressionTests?

src/Analyzer.JsonRuleEngine/JsonRuleEvaluation.cs
```diff
@@ -74,6 +74,26 @@ public bool HasResults
...
+
+        /// <summary>
+        /// Whether or not there are any results or evaluations associated with this <see cref="JsonRuleEvaluation"/>.
+        /// </summary>
+        /// <returns>True if there are no results and no evaluations in this <see cref="JsonRuleEvaluation"/> or a sub-<see cref="JsonRuleEvaluation"/>.
+        /// False otherwise.</returns>
+        public bool NoScopesFound
+        {
+            get => !HasResults && !HasEvaluations;
+        }
```
JohnathonMohr, line 95: Just checking `HasResults` is sufficient.  Actually, checking `HasEvaluations` as well I think is a bug, since there might be nested evaluations without results.

I think `NoScopesFound` isn't needed, since `HasResults` already provides everything needed.

src/Analyzer.JsonRuleEngine/JsonRuleEvaluation.cs
```diff
@@ -74,6 +74,26 @@ public bool HasResults
...
 
+        /// <summary>
+        /// Whether or not there are any evaluations associated with this <see cref="JsonRuleEvaluation"/>.
+        /// </summary>
+        /// <returns>True if there are any evaluations in this <see cref="JsonRuleEvaluation"/>.
+        /// False otherwise.</returns>
+        public bool HasEvaluations
+        {
+            get => Evaluations.Any();
+        }
```
JohnathonMohr, line 85: This likely isn't needed anymore.

src/Analyzer.JsonRuleEngine/Expressions/AllOfExpression.cs
```diff
@@ -43,6 +43,13 @@ public override JsonRuleEvaluation Evaluate(IJsonPathResolver jsonScope)
...
                     var evaluation = expression.Evaluate(scope);
+
+                    // Filter out evaluation if it didn't find the scope to evaluate
+                    if (evaluation.NoScopesFound)
+                    {
+                        continue;
+                    }
+
                     evaluationPassed &= evaluation.Passed;
                     jsonRuleEvaluations.Add(evaluation);
```
JohnathonMohr, line 54: non-blocking: I generally use `continue` when there are large chunks of code I want to skip, or if a condition requiring code to be skipped is deeply nested and non-trivial to arrange differently.  Since this is only skipping two lines of code, I think it would read better having those two lines inside the `if`.

``` suggestion
                    // Only add this Evaluation if something was actually evaluated (Results are present)
                    if (evaluation.HasResults)
                    {
                        evaluationPassed &= evaluation.Passed;
                        jsonRuleEvaluations.Add(evaluation);
                    }
```

JohnathonMohr: I'm also wondering now - should the evaluation still be added to `jsonRuleEvaluations`, but just not be combined into `evaluationPassed`?  Maybe we should have a sync to figure out what we want the behavior to be.

JohnathonMohr (changes requested):

src/Analyzer.JsonRuleEngine.UnitTests/AnyOfExpressionTests.cs
```diff
@@ -88,6 +89,72 @@ public void Evaluate_TwoLeafExpressions_ExpectedResultIsReturned(bool evaluation
             Assert.AreEqual(expectedFalse, anyOfEvaluation.EvaluationsEvaluatedFalse.Count());
         }
 
+        [TestMethod]
+        public void Evaluate_SubResourceScopeNotFound_ExpectedResultIsReturned()
```
VeraBE, line 93: This is also mostly repeated in the tests, right? It probably goes out of the scope of this PR but maybe is worth creating a base class for operators like Any and All, because I'm just realizing those two classes have basically the same code and apparently even some overlapping tests. Should we add an issue for that?

JohnathonMohr: Reducing test code would be nice, but since AllOf and AnyOf only have about 20 lines of significant code, it's hard to imagine how that can be meaningfully reduced further without much effort.  I feel like combining those further by introducing an additional base class wouldn't provide any benefit long term.  What benefit do you see?

VeraBE: I think we would remove more than 20 lines of repeated code since that test alone is 60 and it's probably not the only one overlapping. Besides that, I think it would make the code more clear

JohnathonMohr: I agree.  Combining tests would be very nice.

Combining product code would be nice too, but since there's less than 20 lines of code for the logic of the entire class (for All/AnyOf), I don't know if we would get a good return on the investment in this case.  I could be overestimating the work.  Feel free to create an issue and we can triage it.

yane3628: Created an issue: https://github.com/Azure/template-analyzer/issues/137

src/Analyzer.JsonRuleEngine/Expressions/AllOfExpression.cs
yane3628: Added this as a note to discuss.

src/Analyzer.JsonRuleEngine/Expressions/AllOfExpression.cs
```diff
@@ -43,8 +43,13 @@ public override JsonRuleEvaluation Evaluate(IJsonPathResolver jsonScope)
                 foreach (var expression in AllOf)
                 {
                     var evaluation = expression.Evaluate(scope);
-                    evaluationPassed &= evaluation.Passed;
-                    jsonRuleEvaluations.Add(evaluation);
+
+                    // Add evaluations if scopes where found to evaluate
```
JohnathonMohr, line 47: non-blocking: ... *were* found ... (also in AnyOf)

JohnathonMohr (approved):

VeraBE (approved):

yane3628:
I pressed the wrong button trying to complete the PR 😅

PR #136, opened 2021-06-17
Remove unimplemented commands

VeraBE:
Remove unimplemented commands from the CLI because showing some of those options in "--help" generates confusion


src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -41,10 +41,7 @@ private RootCommand SetupCommandLineAPI()
             rootCommand = new RootCommand();
             rootCommand.Description = "Analyze Azure Resource Manager (ARM) Templates for security and best practice issues.";
 
-            // It has two commands - analyze-template and analyze-directory
             rootCommand.AddCommand(SetupAnalyzeTemplateCommand());
```
JohnathonMohr, line 44: Can you update this too, so instead of adding a new command, it just adds options directly to the root command?

JohnathonMohr (approved):

yane3628 (approved):
@norbusonam FYI some of the code in the CLI is changing

Issue #88, opened 2021-04-16
Add support for having dateTime operators


yane3628:
When doing a dateTime comparison, a rule should be able to add/subtract time from the current time.

Features to support
- addDays(int numberOfDays)
- utcNow()

Related to #52

PR #139, opened 2021-06-22
Support inequality operators

### LINKED ISSUE #89: Add support for greater/lesser operators
Rules should be able to compare numbers and dates with greater/lesser operators

### Examples

```
{
    "path": "outputs.numberOfResourcesDeployed.value",
    "less": 1
}
```

```
{
    "path": "outputs.numberOfResourcesDeployed.value",
    "lessOrEquals": 1
}
```

```
{
    "path": "outputs.numberOfResourcesDeployed.value",
    "greater": 1
}
```

```
{
    "path": "outputs.numberOfResourcesDeployed.value",
    "greaterOrEquals": 1
}
```

```
"evaluation": {
    "resourceType": "Microsoft.KeyVault.Data/vaults/certificates",
    "path": "attribures.expiresOn",
    "lessOrEquals": "01/01/2021"
  }
```

Related to #56, #52

VeraBE:
Closes #89 

There are a couple of TODOs in the code to discuss during the review


src/Analyzer.JsonRuleEngine.UnitTests/ExpressionConverterTests.cs
```diff
@@ -42,6 +42,10 @@ public class ExpressionConverterTests
         [DataRow("equals", "someString", "exists", true, DisplayName = "{\"Equals\": \"someString\"}, Where: {\"Exists\": true}")]
         [DataRow("notEquals", 0, DisplayName = "{\"NotEquals\": 0}")]
         [DataRow("regex", "regexPattern", DisplayName = "{\"Regex\": \"regexPattern\"}")]
+        [DataRow("greater", 100, DisplayName = "{\"Greater\": 100}")]
+        [DataRow("less", 100, DisplayName = "{\"Less\": 100}")]
+        [DataRow("greaterOrEquals", 100, DisplayName = "{\"GreaterOrEquals\": 100}")]
+        [DataRow("lessOrEquals", 100, DisplayName = "{\"LessOrEquals\": 100}")]
```
yane3628, line 48: Add a test for a date time value

src/Analyzer.JsonRuleEngine.UnitTests/InequalityOperatorTests.cs
```diff
@@ -0,0 +1,160 @@
...
+        [DataRow(2.8, 1.3, true, true, false, DisplayName = "A float is not less or equal to another float")]
+        [DataRow(1.3, 1.3, true, true, true, DisplayName = "A float is less or equal to another float because both are equal")]
+        public void EvaluateExpression_ValidNumericType_ReturnsExpectedEvaluationResult(object leftValue, object rightValue, bool isNegative, bool orEquals, bool evaluationResult)
+        {
+            CompareObjects(leftValue, rightValue, isNegative, orEquals, evaluationResult);
+        }
+
+        [DataTestMethod]
+        // >date
+        [DataRow(637676928000000000, 637500672000000000, false, false, true, DisplayName = "A date is greater than another date")]
```
yane3628, line 80: Is this how the date would have to be formatted in the rule?

src/Analyzer.JsonRuleEngine/Operators/InequalityOperator.cs
```diff
@@ -0,0 +1,127 @@
...
+    /// </summary>
+    internal class InequalityOperator : LeafExpressionOperator
+    {
+        /// <inheritdoc/>
+        public override string Name => GetName();
+
+        /// <summary>
+        /// Whether the operator also considers equality
+        /// </summary>
+        public Boolean OrEquals;
```
yane3628, line 21: Can be simplified to type bool

yane3628 (changes requested):

src/Analyzer.JsonRuleEngine/Operators/InequalityOperator.cs
```diff
@@ -0,0 +1,127 @@
...
+
+        /// <summary>
+        /// Creates an InequalityOperator.
+        /// </summary>
+        /// <param name="specifiedValue">The value specified in the JSON rule.</param>
+        /// <param name="isNegative">Whether the operator compares by greater than or by less than.</param>
```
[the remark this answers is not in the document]
JohnathonMohr: Can you rename the `isNegative` parameter to something more descriptive (like `greaterThan`, etc.)?  It can still use the IsNegative class member internally to store it.

src/Analyzer.JsonRuleEngine/Operators/InequalityOperator.cs
```diff
@@ -0,0 +1,127 @@
...
+                return false;
+            }
+
+            ValidateComparisonTerm(tokenToEvaluate);
+
+            // TODO throw this exception earlier?
+            if ((SpecifiedValue.Type == JTokenType.Date && tokenToEvaluate.Type != JTokenType.Date) ||
+                (tokenToEvaluate.Type == JTokenType.Date && SpecifiedValue.Type != JTokenType.Date))
+            {
+                throw new InvalidOperationException($"Cannot compare {SpecifiedValue.Type} with {tokenToEvaluate.Type} using an InequalityOperator");
```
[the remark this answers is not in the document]
JohnathonMohr: This should just evaluate to `false`.

src/Analyzer.JsonRuleEngine/Operators/InequalityOperator.cs
```diff
@@ -0,0 +1,127 @@
...
+            }
+            else if (!IsNegative && OrEquals)
+            {
+                return "GreaterOrEquals";
+            }
+            else
+            {
+                return "Greater";
+            }
+        }
```
JohnathonMohr: This can be simplified a bit and look a little tidier
``` suggestion
        private string GetName() =>
            IsNegative && OrEquals ? "LessOrEquals" :
            IsNegative && !OrEquals ? "Less" :
            !IsNegative && OrEquals ? "GreaterOrEquals" :
            "Greater";
```

src/Analyzer.JsonRuleEngine/Operators/InequalityOperator.cs
```diff
@@ -0,0 +1,132 @@
...
+        /// </summary>
+        /// <param name="specifiedValue">The value specified in the JSON rule.</param>
+        /// <param name="greater">Whether the operator compares by greater than or by less than.</param>
+        /// <param name="orEquals">Whether the operator also considers equality.</param>
+        public InequalityOperator(JToken specifiedValue, bool greater, bool orEquals)
+        {
+            if (specifiedValue == null)
+            {
+                throw new ArgumentNullException(nameof(specifiedValue));
+            }
```
JohnathonMohr, line 39:
``` suggestion
            this.SpecifiedValue = specifiedValue ?? throw new ArgumentNullException(nameof(specifiedValue));
```

src/Analyzer.JsonRuleEngine/Operators/InequalityOperator.cs
```diff
@@ -0,0 +1,132 @@
...
+        /// <returns>A value indicating whether or not the evaluation passed.</returns>
+        public override bool EvaluateExpression(JToken tokenToEvaluate)
+        {
+            if (tokenToEvaluate == null)
+            {
+                // If the specified property in the JSON is not defined then we would assume it could potentially have an undesired value
+                return false;
+            }
+
+            ValidateComparisonTerm(tokenToEvaluate);
```
JohnathonMohr, line 63: It shouldn't throw here either.

src/Analyzer.JsonRuleEngine/Operators/InequalityOperator.cs
```diff
@@ -0,0 +1,132 @@
...
+            }
+            else if (!Greater && OrEquals)
+            {
+                return "LessOrEquals";
+            }
+            else
+            {
+                return "Less";
+            }
+        }
```
JohnathonMohr, line 130: This can be simplified a bit and look tidier :)
``` suggestion
        private string GetName() =>
            Greater && OrEquals ? "GreaterOrEquals" :
            Greater && !OrEquals ? "Greater" :
            !Greater && OrEquals ? "LessOrEquals" :
            "Less";
```

src/Analyzer.JsonRuleEngine/Operators/InequalityOperator.cs
```diff
@@ -0,0 +1,132 @@
...
+            {
+                value = token.Value<DateTime>().ToOADate();
+            }
+            else
+            {
+                value = token.Value<double>();
+            }
+
+            return value;
+        }
```
JohnathonMohr, line 111:
``` suggestion
        private double GetNormalizedValue(JToken token) =>
            token.Type == JTokenType.Date
                ? token.Value<DateTime>().ToOADate()
                : token.Value<double>();
```

src/Analyzer.JsonRuleEngine/Operators/InequalityOperator.cs
```diff
@@ -0,0 +1,127 @@
...
+            {
+                result = result || normalizedSpecifiedValue == normalizedTokenToEvaluate;
+
+            }
+
+            return result;
+        }
+
+        // TODO throw this exception earlier?
+        private void ValidateComparisonTerm(JToken term)
```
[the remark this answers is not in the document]
JohnathonMohr: I think as long as the validation happens during the "rule interpretation" phase and before evaluating a template, it should be ok.  This would fall in the "interpretation" phase (creating the rule) so I think it's fine.

src/Analyzer.JsonRuleEngine/Operators/InequalityOperator.cs
```diff
@@ -0,0 +1,132 @@
...
+
+            var normalizedSpecifiedValue = GetNormalizedValue(SpecifiedValue);
+            var normalizedTokenToEvaluate = GetNormalizedValue(tokenToEvaluate);
+
+            var result = Greater ? normalizedSpecifiedValue > normalizedTokenToEvaluate : normalizedSpecifiedValue < normalizedTokenToEvaluate;
+
+            if (OrEquals)
+            {
+                result = result || normalizedSpecifiedValue == normalizedTokenToEvaluate;
+
```
JohnathonMohr, line 80: \<minor> extra newline

src/Analyzer.JsonRuleEngine/Operators/InequalityOperator.cs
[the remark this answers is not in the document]
VeraBE: Why do you think this should evaluate to false? To avoid stopping the report of other results like Yanelis said? I'm more inclined to keep throwing an exception and catch it outside of the operator to show the information in the console but continue with the analysis, to give more information to the user, and also because returning false seems wrong in the context of the class. The Regex operator also throws an exception when trying to process an invalid regex. Should I create an issue to catch these exceptions?

src/Analyzer.JsonRuleEngine.UnitTests/ExpressionConverterTests.cs
```diff
@@ -42,6 +42,10 @@ public class ExpressionConverterTests
...
         [DynamicData(nameof(InOperatorTestScenarios), DynamicDataDisplayName = nameof(GetInOperatorTestDisplayName))]
         public void ReadJson_LeafWithValidOperator_ReturnsCorrectTypeAndValues(string operatorProperty, object operatorValue, params object[] whereCondition)
```
JohnathonMohr, line 50: Add tests below in the invalid cases too.

src/Analyzer.JsonRuleEngine.UnitTests/InequalityOperatorTests.cs
VeraBE: No no, I wrote the tests like that because it seemed easier

`   if (token.Type == JTokenType.Date) 
            {
                value = token.Value<DateTime>().ToOADate();
             }`
             
Any date format recognized by JToken as a Date should work. That's why I added this TODO: `// TODO test and document JSON date format?`. At the time I thought we might want to discuss this but I think we really should add it to the documentation, because JSON doesn't define a date format, doing it now

src/Analyzer.JsonRuleEngine.UnitTests/InequalityOperatorTests.cs
```diff
@@ -0,0 +1,160 @@
...
+        [TestMethod]
+        public void GetName_ReturnsCorrectName()
+        {
+            var specifiedValue = TestUtilities.ToJToken(100);
+
+            Assert.AreEqual("LessOrEquals", new InequalityOperator(specifiedValue, greater: false, orEquals: true).Name);
+            Assert.AreEqual("Less", new InequalityOperator(specifiedValue, greater: false, orEquals: false).Name);
+            Assert.AreEqual("GreaterOrEquals", new InequalityOperator(specifiedValue, greater: true, orEquals: true).Name);
+            Assert.AreEqual("Greater", new InequalityOperator(specifiedValue, greater: true, orEquals: false).Name);
+
```
JohnathonMohr, line 37: \<minor> extra newline

src/Analyzer.JsonRuleEngine.UnitTests/InequalityOperatorTests.cs
```diff
@@ -0,0 +1,160 @@
...
+        [DataRow(2.8, 2.8, false, false, false, DisplayName = "A float is not less than another float because both are equal")]
+        // >=float
+        [DataRow(2.8, 1.3, true, true, true, DisplayName = "A float is greater or equal to another float")]
+        [DataRow(1.3, 2.8, true, true, false, DisplayName = "A float is not greater or equal to another float")]
+        [DataRow(2.8, 2.8, true, true, true, DisplayName = "A float is greater or equal to another float because both are equal")]
+        // <=float
+        [DataRow(1.3, 2.8, false, true, true, DisplayName = "A float is less or equal to another float")]
+        [DataRow(2.8, 1.3, false, true, false, DisplayName = "A float is not less or equal to another float")]
+        [DataRow(1.3, 1.3, false, true, true, DisplayName = "A float is less or equal to another float because both are equal")]
+        public void EvaluateExpression_ValidNumericType_ReturnsExpectedEvaluationResult(object leftValue, object rightValue, bool greater, bool orEquals, bool evaluationResult)
```
JohnathonMohr, line 73: Add a few combinations of comparing int and float.

src/Analyzer.JsonRuleEngine.UnitTests/InequalityOperatorTests.cs
```diff
@@ -0,0 +1,160 @@
...
+        // >=date
+        [DataRow(637676928000000000, 637500672000000000, true, true, true, DisplayName = "A date is greater or equal to another date")]
+        [DataRow(637500672000000000, 637676928000000000, true, true, false, DisplayName = "A date is not greater or equal to another date")]
+        [DataRow(637676928000000000, 637676928000000000, true, true, true, DisplayName = "A date is greater or equal to another date because both are equal")]
+        // <=date
+        [DataRow(637500672000000000, 637676928000000000, false, true, true, DisplayName = "A date is less or equal to another date")]
+        [DataRow(637676928000000000, 637500672000000000, false, true, false, DisplayName = "A date is not less or equal to another date")]
+        [DataRow(637500672000000000, 637500672000000000, false, true, true, DisplayName = "A date is less or equal to another date because both are equal")]
+        public void EvaluateExpression_ValidDateType_ReturnsExpectedEvaluationResult(long leftTicks, long rightTicks, bool greater, bool orEquals, bool evaluationResult)
```
JohnathonMohr, line 95: There are functional parsing tests in the project Analyzer.JsonRuleEngine.FunctionalTests in RuleParsingTests.cs.  I think it would be valuable to add some tests parsing at least date values in this operator.

JohnathonMohr (changes requested):

src/Analyzer.JsonRuleEngine/Operators/InequalityOperator.cs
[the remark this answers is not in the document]
JohnathonMohr: The Regex operator throws when interpreting the rule definition (the regex defined in the rule).  This is important because the rule can't evaluate anything if it's not defined properly.  That is different than throwing while evaluating a template, which is what's happening here.

Regardless of whether the two values are technically comparable or not, the intent is to simply discover places in a template that are inadequate or need to change.  If they aren't comparable, simply returning `false` will result in a rule failure, and the user will be notified of the failure and the line to look at.  This is cleaner and easier to understand than throwing an exception that needs to be caught later.

src/Analyzer.JsonRuleEngine.UnitTests/InequalityOperatorTests.cs
JohnathonMohr: I added a comment about this in the tests too.  You can add functional parsing tests in Analyzer.JsonRuleEngine.FunctionalTests in RuleParsingTests.cs.

src/Analyzer.JsonRuleEngine/Operators/InequalityOperator.cs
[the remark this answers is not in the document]
VeraBE: But If the rule is written negating the output of this operator in some way? The user could potentially not find out about the issue. Or if the inequality operator restricts the search space like: where value > 10 check something else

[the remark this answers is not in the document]
yane3628: It could also be that the rule is expecting an int or something, but that property is never going to be an int

src/Analyzer.JsonRuleEngine/Operators/InequalityOperator.cs
VeraBE: I didn't commit the suggestion because I rather not define the function if we are doing it this way, but made the change

src/Analyzer.JsonRuleEngine.UnitTests/InequalityOperatorTests.cs
VeraBE: Thank you! I wasn't sure where to add those tests

src/Analyzer.JsonRuleEngine.UnitTests/InequalityOperatorTests.cs
VeraBE: While doing this I realized there might be a bigger issue. I didn't realize there are many ways to specify dates in ARM templates, but JToken doesn't seem to parse all of them :( Working on that

src/Analyzer.JsonRuleEngine/Operators/InequalityOperator.cs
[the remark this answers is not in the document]
JohnathonMohr: @VeraBE, @yane3628, those are good examples where we might want something more than a simple pass/fail.  I think this does deserve more thought on how we handle these scenarios, and what information to surface to the user.

However, I don't think throwing an exception here is the right way to do it, and I don't think there's a pressing need to make issues like this highly visible just yet.  This should be designed in a more holistic way before implementing something like this.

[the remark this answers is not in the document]
VeraBE: Okay, do you want me to add an issue for that? If we are going to return false for cases like that in the meantime, I feel the issue should have a pretty high priority, because we are risking many false negatives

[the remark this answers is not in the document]
JohnathonMohr: Yes, go ahead and create an issue for this.

src/Analyzer.JsonRuleEngine.UnitTests/InequalityOperatorTests.cs
VeraBE: @yane3628 @JohnathonMohr I'm not sure how to handle dates specified using [these functions](https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-date). Should I add support for that in this PR? Do we want to add support for dates comparisons without handling those cases?

I also haven't been able to find a complete documentation on date formates supported by ARM templates, besides 2021-02-28T18:17:16.543Z and 2021-02-28, I assume supporting only those two formats is not enough, right?

yane3628: dateTimeAdd and other functions are listed as another issue https://github.com/Azure/template-analyzer/issues/88.

The docs also mention this format 20190305T175318Z

src/Analyzer.JsonRuleEngine.UnitTests/ExpressionConverterTests.cs
VeraBE: I'm not sure if this is an outdated comment or I'm not seeing the missing test cases

src/Analyzer.JsonRuleEngine.UnitTests/InequalityOperatorTests.cs
VeraBE: I went through all (😵) the quick-start templates, and the documentation of the resources I found that specify dates, as Johnathon suggested, and came up with the accepted formats in the documentation

I also used [this ](https://docs.microsoft.com/en-us/rest/api/storageservices/formatting-datetime-values) as a guide of how to express it

I know the [date functions](https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-date) allow a ton more formats but I suppose that's out of scope for this PR

src/Analyzer.JsonRuleEngine.UnitTests/InequalityOperatorTests.cs
```diff
@@ -101,63 +101,39 @@ public void GetName_ReturnsCorrectName()
...
+        [DataRow("2021-09-20", "2021-09-20", true, true, true, DisplayName = "A date is greater or equal to another date because both are equal")]
         // <=date
-        [DataRow(637500672000000000, 637676928000000000, false, true, true, DisplayName = "A date is less or equal to another date")]
-        [DataRow(637676928000000000, 637500672000000000, false, true, false, DisplayName = "A date is not less or equal to another date")]
-        [DataRow(637500672000000000, 637500672000000000, false, true, true, DisplayName = "A date is less or equal to another date because both are equal")]
-        public void EvaluateExpression_ValidDateType_ReturnsExpectedEvaluationResult(long leftTicks, long rightTicks, bool greater, bool orEquals, bool evaluationResult)
+        [DataRow("2021-02-28", "2021-09-20", false, true, true, DisplayName = "A date is less or equal to another date")]
+        [DataRow("2021-09-20", "2021-02-28", false, true, false, DisplayName = "A date is not less or equal to another date")]
+        [DataRow("2021-02-28", "2021-02-28", false, true, true, DisplayName = "A date is less or equal to another date because both are equal")]
```
yane3628, line 119: Add tests for the other supported formats. At least one for each.

yane3628 (approved):
Small comments

src/Analyzer.JsonRuleEngine.UnitTests/InequalityOperatorTests.cs
VeraBE: I added test cases like that in RuleParsingTests, do you think those should be "moved" here or it makes sense to have both?

yane3628: I think there should still be a test here to show that that format actually gets evaluated correctly.

src/Analyzer.JsonRuleEngine/Operators/InequalityOperator.cs
```diff
@@ -0,0 +1,103 @@
...
+            if (term.Type == JTokenType.String)
+            {
+                try
+                {
+                    return DateTime.Parse(term.Value<string>(), styles: DateTimeStyles.RoundtripKind).ToOADate();
+                }
+                catch
+                {
+                    return null;
+                }
```
JohnathonMohr, line 93: Maybe this?
``` suggestion
                return DateTime.TryParse(term.Value<string>(), provider: null, styles: DateTimeStyles.RoundtripKind, out var date)
                    ? date.ToOADate()
                    : null;
```

JohnathonMohr (changes requested):

src/Analyzer.JsonRuleEngine.UnitTests/ExpressionConverterTests.cs
VeraBE: I tried but at that stage it doesn't fail with invalid values like that

JohnathonMohr: Oh, that makes sense, I guess those tests would be in the operator tests, not in Expression conversion.

src/Analyzer.JsonRuleEngine/Operators/InequalityOperator.cs
VeraBE: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs8400?f1url=%3FappId%3Droslyn%26k%3Dk(CS8400) :/

src/Analyzer.JsonRuleEngine.UnitTests/InequalityOperatorTests.cs
VeraBE: Thank you! It made me catch an error

src/Analyzer.JsonRuleEngine/Operators/InequalityOperator.cs
JohnathonMohr: Oh - `null` just needs to be cast to a `double?`.
``` suggestion
                return DateTime.TryParse(term.Value<string>(), provider: null, styles: DateTimeStyles.RoundtripKind, out var date)
                    ? date.ToOADate()
                    : (double?)null;
```

JohnathonMohr (approved):

PR #140, opened 2021-06-22
Make AnyOf not fail if nothing is evaluated

### LINKED ISSUE #138: AnyOf Operator should not fail if no Expressions were evaluated
If all Expressions defined in an AnyOf operator are skipped due to not finding relevant scopes (i.e. wildcard in path not matching anything, resource type not found), the AnyOf operator should "pass" with no results.

Currently, AnyOf fails in this scenario.

JohnathonMohr:
Updates AnyOf operator to only fail if something was evaluated and something internal failed.  Otherwise, if nothing is evaluated, pass by default.  This ensures rules in which no paths/resource types are relevant inside an AnyOf don't result in a failure.  Closes #138.

- Update AnyOf with bug fix
- Update corresponding test for AnyOf and AllOf

### DIFF
--- a/src/Analyzer.JsonRuleEngine/Expressions/AnyOfExpression.cs
+++ b/src/Analyzer.JsonRuleEngine/Expressions/AnyOfExpression.cs
@@ -39,7 +39,6 @@ namespace Microsoft.Azure.Templates.Analyzer.RuleEngines.JsonEngine.Expressions
             return EvaluateInternal(jsonScope, scope =>
             {
                 List<JsonRuleEvaluation> jsonRuleEvaluations = new List<JsonRuleEvaluation>();
-                bool evaluationPassed = false;
 
                 foreach (var expression in AnyOf)
                 {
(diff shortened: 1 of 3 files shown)


src/Analyzer.JsonRuleEngine/Expressions/AnyOfExpression.cs
```diff
@@ -48,11 +47,10 @@ public override JsonRuleEvaluation Evaluate(IJsonPathResolver jsonScope)
                     // Add evaluations if scopes were found to evaluate
                     if (evaluation.HasResults)
                     {
-                        evaluationPassed |= evaluation.Passed;
                         jsonRuleEvaluations.Add(evaluation);
                     }
                 }
-
+                bool evaluationPassed = jsonRuleEvaluations.Count == 0 || !jsonRuleEvaluations.TrueForAll(e => !e.Passed);
```
VeraBE, line 53:
```suggestion
                bool evaluationPassed = jsonRuleEvaluations.Count == 0 || jsonRuleEvaluations.Any(e => e.Passed);
```

yane3628 (approved):

VeraBE (approved):

PR #143, opened 2021-06-24
Small CLI output improvements

yane3628:
- Print all inner exception messages
- Only print line numbers when the rule failed

### DIFF
--- a/src/Analyzer.Cli/CommandLineParser.cs
+++ b/src/Analyzer.Cli/CommandLineParser.cs
@@ -99,8 +99,21 @@ namespace Microsoft.Azure.Templates.Analyzer.Cli
             }
             catch (Exception exp)
             {
-                Console.WriteLine($"An exception occured: {exp.Message}");
+                Console.WriteLine($"An exception occured: {GetAllExceptionMessages(exp)}");
             }
+        }
+
+        private static string GetAllExceptionMessages(Exception exception)
+        {
+            string exceptionMessage = exception.Message;
+
+            while (exception.InnerException != null)
+            {
+                exception = exception.InnerException;
+                exceptionMessage += " " + exception.Message;
+            }
+
+            return exceptionMessage;
         }
 
         private string GenerateResultString(Types.IEvaluation evaluation)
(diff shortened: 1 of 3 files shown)


VeraBE (approved):

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -99,10 +99,23 @@ private void RootCommandHandler(FileInfo templateFilePath, FileInfo parametersFi
...
         }
 
+        private static string GetAllExceptionMessages(Exception exception)
+        {
+            string exceptionMessage = exception.Message;
+
+            while (exception.InnerException != null)
+            {
+                exception = exception.InnerException;
+                exceptionMessage += " " + exception.Message;
```
JohnathonMohr, line 113: Can we use a different separating character besides a single space?

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -111,7 +124,10 @@ private string GenerateResultString(Types.IEvaluation evaluation)
             {
                 foreach (var result in evaluation.Results)
                 {
-                    resultString += $"{TwiceIndentedNewLine}Line: {result.LineNumber}";
+                    if (!result.Passed)
+                    {
+                        resultString += $"{TwiceIndentedNewLine}Line: {result.LineNumber}";
+                    }
```
JohnathonMohr, line 130: This message is already not printed if the evaluation passed (line 85).  Instead of checking here, I suggest moving the call to `GenerateResultString` inside the `if` statement on line 85.

src/Analyzer.Cli/CommandLineParser.cs
yane3628: @JohnathonMohr Any suggestions? I can add a new line or maybe "Inner Exception: ".

JohnathonMohr: Maybe a "|" pipe character?  Or surround messages in square brackets?  The issue with a space is it can make distinguishing distinct messages difficult, like if a message doesn't end with a period (or has multiple sentences with periods) - where one ends and other starts could be ambiguous and make the overall message hard to interpret.

As long as we know messages are easy to distinguish, *how* they're separated isn't terribly important.  It just needs to be *easy* :)

JohnathonMohr:
Was the inclusion of rule updates intentional?

yane3628:
> Was the inclusion of rule updates intentional?

@JohnathonMohr Yea, they were small changes and I didn't want to make a whole new PR for.

JohnathonMohr (approved):

PR #151, opened 2021-07-08
Generalize XML comment enforcement

JohnathonMohr:
- Putting property group that generates code documentation (and enforces XML comments) into common *Directory.Build.targets* file
  - Only applies to projects that aren't tests
- Removing from all projects that declared them manually

### DIFF
--- a/src/Analyzer.Cli/Analyzer.Cli.csproj
+++ b/src/Analyzer.Cli/Analyzer.Cli.csproj
@@ -7,12 +7,6 @@
     <Description>A command line interface for Microsoft.Azure.Templates.Analyzer.Core - an ARM template scanner for security misconfigurations and best practices</Description>
   </PropertyGroup>
 
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-    <WarningsAsErrors />
-    <DocumentationFile>bin\$(AssemblyName).xml</DocumentationFile>
-  </PropertyGroup>
-
   <ItemGroup>
     <PackageReference Include="System.CommandLine" Version="2.0.0-beta1.20574.7" />
   </ItemGroup>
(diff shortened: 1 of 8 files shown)


yane3628 (approved):

VeraBE (commented):
We don't need to import the new targets file, or something related, to the project files?

JohnathonMohr:
@VeraBE Directory.Build.props and .targets are special cases that are imported automatically.  https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2019#directorybuildprops-and-directorybuildtargets

VeraBE:
> @VeraBE Directory.Build.props and .targets are special cases that are imported automatically. https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2019#directorybuildprops-and-directorybuildtargets

Yes. but I read in a few places that we need to import Microsoft.Common.targets somehow for that to work, right? (GitHub is not letting me reply in the same conversation :O)

JohnathonMohr:
@VeraBE that was true prior to MSBuild version 15 (noted in the link).  (Comments at the PR level [not on a line in a file] are all grouped in a single thread in GH. ☹)

VeraBE:
@JohnathonMohr The first line of the documentation says "MSBuild projects that use the standard build process (importing Microsoft.Common.props and Microsoft.Common.targets) have several extensibility hooks that you can use to customize your build process.". Do you interpret that as the standard build process makes those imports? I checked the repo and couldn't find imports like that

JohnathonMohr:
> @JohnathonMohr The first line of the documentation says "MSBuild projects that use the standard build process (importing Microsoft.Common.props and Microsoft.Common.targets) have several extensibility hooks that you can use to customize your build process.". Do you interpret that as the standard build process makes those imports? I checked the repo and couldn't find imports like that

@VeraBE I was just noticing that too.  I'm guessing building with `dotnet` (.NET Core and .NET 5+) do that as part of the default build process now without needing to explicitly include it in the project files (kind of like how all .cs files at the same or sub-directory are automatically included and compiled in a project file).  Still looking to verify this - but in the meantime, this does work one way or another, just not certain on the technical *how* yet.

VeraBE (approved):

JohnathonMohr:
@VeraBE I think this is where the magic comes from: https://docs.microsoft.com/en-us/dotnet/core/project-sdk/overview

PR #164, opened 2021-07-15
Add SQL_Retention_Days built in rule and fix related bugs

### LINKED ISSUE #56: [Built-in Rule] SQL servers should retain audit data for at least 90 days
Azure policy: https://github.com/Azure/azure-policy/blob/master/built-in-policies/policyDefinitions/SQL/SqlServerAuditingRetentionDays_Audit.json

``` javascript
{
  "name": "SQL_Rentention_Days",
  "description": "SQL servers with auditing to storage account destination should be configured with 90 days retention or higher",
  "recommendation": "Set the data retention for your SQL Server's auditing to storage account destination to at least 90 days",
  "helpUri": "https://github.com/Azure/template-analyzer/docs/built-in-bpa-rules.md/#sql-servers-with-auditing-to-storage-account-destination-should-be-configured-with-90-days-retention-or-higher",
  "evaluation": {
    "resourceType": "Microsoft.Sql/servers",
    "where": {
      "path": "properties.kind",
      "regex": "^((?!analytics).)*$"
    },
    "evaluate": {
      "resourceType": "Microsoft.Sql/servers/auditingSettings",
      "where": {
        "path": "name",
        "equals": "default"
      },
      "allOf": [
        {
          "anyOf": [
            {
              "path": "properties.isAzureMonitorTargetEnabled",
              "equals": false
            },
            {
              "path": "properties.storageEndpoint",
              "hasValue": true
            }
          ]
        },
        {
          "path": "properties.retentionDays",
          "greaterOrEqual": 90
        }
      ]
    }
  }
}
```

### LINKED ISSUE #159: Inequality operators are interpreted opposite as intended
If a rule specifies:

```
{
    "path": "properties.retentionDays",
    "greaterOrEquals": 90
}
```

an evaluation will fail if the template has a value greater than 90. It should behave in the opposite way

### LINKED ISSUE #160: BuiltInRuleTests reports failed partial evaluations
When a rule fails for a case but succeeds for others, the tests are expecting reported failures also for the failed partial evaluations of the ones that succeed

VeraBE:
Closes #56
Closes #159 
Closes #160


yane3628:
@VeraBE For the issue linking automation to work you need to explicitly add the work "Closes" for each linked issue. Or you can manually link the issues on the right. Notice only issue 56 is linked even though you mention the other two in the description.

src/Analyzer.Core.BuiltInRuleTests/Tests/SQL_Rentention_Days.json
```diff
@@ -0,0 +1,22 @@
...
+        "LineNumber": 127,
+        "Description": "isAzureMonitorTargetEnabled set to false"
+      },
+      {
+        "LineNumber": 128,
+        "Description": "storageEndpoint is not an empty string"
+      },
+      {
+        "LineNumber": 129,
+        "Description": "retentionDays set to a value different than 0 and less than 90"
```
nonik0, line 18: I think this could be phrased differently? Would be it more clear to say that retentionDays was set to a value greater than 0 and less than 90? I guess this wouldn't work if negative retention days is a valid template value though, do you know if that would even pass validation?

src/Analyzer.Core.BuiltInRuleTests/Tests/SQL_Rentention_Days.json
```diff
@@ -0,0 +1,22 @@
+﻿[
```
nonik0, line 1: nit: typo in filename

src/Analyzer.JsonRuleEngine.UnitTests/InequalityOperatorTests.cs
```diff
@@ -145,9 +145,9 @@ private void CompareObjects(object left, object right, bool greater = false, boo
             var leftToken = TestUtilities.ToJToken(left);
             var rightToken = TestUtilities.ToJToken(right);
 
-            var inequalityOperator = new InequalityOperator(leftToken, greater: greater, orEquals: orEquals);
+            var inequalityOperator = new InequalityOperator(rightToken, greater: greater, orEquals: orEquals);
```
nonik0, line 148: I don't understand this change, it seems like you're flipping it around so I think I am missing something? Why is the value given to the constructor operating on as the right hand operator for the comparison? I feel like it makes more sense how it already is?

src/Analyzer.Core.BuiltInRuleTests/Tests/SQL_Rentention_Days.json
VeraBE: I wrote it like that because the rule specifies that retentionDays should equal to 0 or be greater than 90. Considering that, do you think is still confusing?

src/Analyzer.Core.BuiltInRuleTests/Tests/SQL_Rentention_Days.json
VeraBE: Oh no, the typo is everywhere

src/Analyzer.JsonRuleEngine.UnitTests/InequalityOperatorTests.cs
VeraBE: I confused the value in the template that the operator checks against with the value in the rule, here's an example: https://github.com/Azure/template-analyzer/issues/159

src/Analyzer.Core.BuiltInRuleTests/Tests/SQL_Retention_Days.json
```diff
@@ -0,0 +1,22 @@
+﻿[
+  {
+    "Template": "SQLServerAuditingSettings",
+    "ReportedFailures": [
+      // The resource between lines 10 and 33 passes this check because all the elements of the AnyOf are true
+      // The ones between lines 34 and 113 passes because at least one of the elements of the AnyOf is true
+      // And the ones between lines 135 and 173 passes because of empty scopes
```
JohnathonMohr, line 7: Can you explain the last scenario a bit more?  The idea of "empty scopes" is more of a development concept, and wouldn't make as much sense to someone just creating rules (and reviewing existing tests).  From a rule author perspective, the resources pass because their property values exclude them from the rule criteria.  (You could maybe say the resources are "out of scope" for X/Y reasons, but "scope" in that phrase would have a slightly different meaning and would therefore make sense to a rule author.)

src/Analyzer.JsonRuleEngine.UnitTests/InequalityOperatorTests.cs
JohnathonMohr: I thought this was confusing at first too, but now I think this is actually a better representation.

If a rule is written like:
```
{
    "path": "property.value"
    "lessThan": 50
}
```
... the concept being expressed is `'property.value' < 50`, in which case the `InequalityOperator` should actually be initialized with the right token.

src/Analyzer.JsonRuleEngine/Operators/InequalityOperator.cs
```diff
@@ -69,7 +69,7 @@ public override bool EvaluateExpression(JToken tokenToEvaluate)
                 return false; // Not ideal, will be improved in the future
             }
 
-            var result = Greater ? EffectiveValue > finalTokenToEvaluate : EffectiveValue < finalTokenToEvaluate;
+            var result = Greater ? EffectiveValue < finalTokenToEvaluate : EffectiveValue > finalTokenToEvaluate;
```
JohnathonMohr, line 72: Can you rearrange these tokens so `Greater` actually uses the `>` operator?

src/Analyzer.Core/Rules/BuiltInRules.json
```diff
@@ -663,5 +663,49 @@
         }
       ]
     }
+  },
+  {
+    "name": "SQL_Retention_Days",
+    "description": "SQL servers with auditing to storage account destination should be configured with 90 days retention or higher",
+    "recommendation": "Set the data retention for your SQL Server's auditing to storage account destination to at least 90 days",
+    "helpUri": "https://github.com/Azure/template-analyzer/docs/built-in-bpa-rules.md/#sql-servers-with-auditing-to-storage-account-destination-should-be-configured-with-90-days-retention-or-higher",
```
JohnathonMohr, line 671: missing '/blob/main` in link

docs/built-in-bpa-rules.md
```diff
@@ -159,4 +159,9 @@ You should only use built-in roles instead of cutom RBAC roles. Custom RBAC role
...
+**Recommendation**: To [use HTTPS to ensure server/service authentication and protect data in transit from network layer eavesdropping attacks](https://docs.microsoft.com/en-us/azure/app-service/configure-ssl-bindings#enforce-https), in the [Microsoft.Web/Sites resource properties](https://docs.microsoft.com/en-us/azure/templates/microsoft.web/sites?tabs=json#siteproperties-object), add (or update) the *httpsOnly* property, setting its value to `true`.
+
+## SQL servers with auditing to storage account destination should be configured with 90 days retention or higher
+Set the data retention for your SQL Server's auditing to storage account destination to at least 90 days.
+
+**Recommendation**: For incident investigation purposes, we recommend setting the data retention for your [SQL Server' auditing to storage account destination](https://docs.microsoft.com/en-us/azure/templates/microsoft.sql/2020-11-01-preview/servers/auditingsettings?tabs=json#serverblobauditingpolicyproperties-object) to at least 90 days. Confirm that you are meeting the necessary retention rules for the regions in which you are operating. This is sometimes required for compliance with regulatory standards.
```
JohnathonMohr, line 167: Can you reword the first sentence a bit?  The statement being made doesn't really say what a user should do at the page being linked to.  (Also, I think there's an 's' missing after the apostrophe in the link text.)

src/Analyzer.Core.BuiltInRuleTests/TestRunner.cs
```diff
@@ -46,9 +46,14 @@ public void TestRule(TestConfiguration ruleExpectations)
...
-                var failingLines = GetAllResults(thisRuleEvaluation)
-                    .Where(r => !r.Passed)
-                    .Select(r => r.LineNumber).ToHashSet();
+                var failingLines = thisRuleEvaluation.Evaluations
+                    .Where(e => !e.Passed)
+                    .Select(e => GetAllResults(e))
+                    .Select(rs => rs.Where(r => !r.Passed)
+                                    .Select(r => r.LineNumber)
+                                    .ToList())
+                    .SelectMany(x => x)
```
JohnathonMohr, line 55: Instead of this simple lambda, change the above `Select` to `SelectMany`.

src/Analyzer.Core/Rules/BuiltInRules.json
```diff
@@ -663,5 +663,49 @@
...
+          "anyOf": [
+            {
+              "allOf": [
+                {
+                  "path": "properties.isAzureMonitorTargetEnabled",
+                  "equals": true
+                },
+                {
+                  "path": "properties.storageEndpoint",
+                  "equals": ""
```
JohnathonMohr, line 694: This might need to be changed to `"hasValue": false`.  The policy is looking for an empty string, but if a template is deployed without this field specified, it probably ends up being an empty string when the policy is evaluated.  We should account for that in the rule.

JohnathonMohr (changes requested):

JohnathonMohr (approved):

src/Analyzer.Core.BuiltInRuleTests/Tests/SQL_Rentention_Days.json
nonik0: Yeah I think "equal to 0 or greater than 90" sounds more clear than the inverse "different than 0 or less than 90" because there's some implicit overlap between different than 0 and less than 90, whereas the conditions equal to 0 or greater than 90 overlap. At least that's how my mind processes it? So imo I think "retentionDays **not** set to a value equal to 0 or greater than 90" is easier to comprehend, also since it's the simple negation of what the expectation is.

nonik0: There's technically some wiggle room to argue that "retentionDays not set to a value equal to 0 or greater than 90" means:
!(retentionDays == 0) || retentionDays > 90
but I still read it as 
!(retentionDays == 0 || retentionDays > 90)

yane3628 (changes requested):

docs/built-in-bpa-rules.md
```diff
@@ -159,4 +159,9 @@ You should only use built-in roles instead of cutom RBAC roles. Custom RBAC role
...
+
+## SQL servers with auditing to storage account destination should be configured with 90 days retention or higher
+Set the data retention for your SQL Server's auditing to storage account destination to at least 90 days.
+
+**Recommendation**: For incident investigation purposes, we recommend setting the data retention for your SQL Server's auditing to storage account destination to at least 90 days, in the [Microsoft.Sql/servers/auditingSettings resource properties](https://docs.microsoft.com/en-us/azure/templates/microsoft.sql/2020-11-01-preview/servers/auditingsettings?tabs=json#serverblobauditingpolicyproperties-object), using the *retentionDays* property. Confirm that you are meeting the necessary retention rules for the regions in which you are operating. This is sometimes required for compliance with regulatory standards.
```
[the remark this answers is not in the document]
VeraBE: I spent like an hour looking around and couldn't find a good Microsoft documentation on that :/

docs/built-in-bpa-rules.md
```diff
@@ -159,4 +159,9 @@ You should only use built-in roles instead of cutom RBAC roles. Custom RBAC role
...
 
-**Recommendation**: To [use HTTPS to ensure server/service authentication and protect data in transit from network layer eavesdropping attacks](https://docs.microsoft.com/en-us/azure/app-service/configure-ssl-bindings#enforce-https), in the [Microsoft.Web/Sites resource properties](https://docs.microsoft.com/en-us/azure/templates/microsoft.web/sites?tabs=json#siteproperties-object), add (or update) the *httpsOnly* property, setting its value to `true`.
\ No newline at end of file
+**Recommendation**: To [use HTTPS to ensure server/service authentication and protect data in transit from network layer eavesdropping attacks](https://docs.microsoft.com/en-us/azure/app-service/configure-ssl-bindings#enforce-https), in the [Microsoft.Web/Sites resource properties](https://docs.microsoft.com/en-us/azure/templates/microsoft.web/sites?tabs=json#siteproperties-object), add (or update) the *httpsOnly* property, setting its value to `true`.
+
+## SQL servers with auditing to storage account destination should be configured with 90 days retention or higher
```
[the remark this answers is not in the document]
VeraBE: Sure, I also added that requirement and a couple more missing to CONTRIBUTING.md

src/Analyzer.Core.BuiltInRuleTests/TestTemplates/SQLServerAuditingSettings.badtemplate
```diff
@@ -0,0 +1,194 @@
...
+      "name": "ASQLServerName",
+      "apiVersion": "2020-02-02-preview",
+      "properties": {
+        "kind": "v12.0"
+      },
+      "resources": [
+        {
+          "apiVersion": "2020-02-02-preview",
+          "type": "Microsoft.Sql/servers/auditingSettings",
+          "name": "default",
```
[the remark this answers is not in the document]
VeraBE: Regarding the simplification, I tried to do that, only the first resource has all the auditingSettings properties, in case this template can be reused later for another rule, and the rest have only the properties we test on, or is there something else I could safely remove?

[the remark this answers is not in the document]
VeraBE: And about the names, I'm inclined to leave the first resource with a generic name for the template reuse, is that okay? I'm renaming the rest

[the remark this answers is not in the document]
JohnathonMohr: You could probably remove `"apiVersion"` from the resources, but it's not a big reduction, so it's fine either way.

yane3628 (approved):

PR #166, opened 2021-07-21
Add not operator support

### LINKED ISSUE #49: [Built-in Rule] Kubernetes Services should be upgraded to a non-vulnerable Kubernetes version
Azure policy: https://github.com/Azure/azure-policy/blob/master/built-in-policies/policyDefinitions/Security%20Center/ASC_UpgradeVersion_KubernetesService_Audit.json

``` javascript
{
    "name": "Kubernetes_Upgrade_Non-Vulnerable_Version",
    "description": "Kubernetes Services should be upgraded to a non-vulnerable Kubernetes version",
    "recommendation": "Upgrade to a non-vulnerable Kubernetes version",
    "helpUri": "https://github.com/Azure/template-analyzer/docs/built-in-bpa-rules.md/#kubernetes-services-should-be-upgraded-to-a-non-vulnerable-kubernetes-version",
    "evaluation": {
        "resourceType": "Microsoft.ContainerService/managedClusters",
        "allOf": [
            {
                "not": {
                    "path": "properties.kubernetesVersion",
                    "in": [
                        "1.13.4",
                        "1.13.3",
                        "1.13.2",
                        "1.13.1",
                        "1.13.0"
                    ]
                }
             },
            {
                "not": {
                    "path": "properties.kubernetesVersion",
                    "in": [
                        "1.12.6",
                        "1.12.5",
                        "1.12.4",
                        "1.12.3",
                        "1.12.2",
                        "1.12.1",
                        "1.12.0"
                    ]
                }
            },
            {
                "not": {
                    "path": "properties.kubernetesVersion",
                    "in": [
                        "1.11.8",
                        "1.11.7",
                        "1.11.6",
                        "1.11.5",
                        "1.11.4",
                        "1.11.3",
                        "1.11.2",
                        "1.11.1",
                        "1.11.0"
                    ]
                }
            },
            {
                "not": {
                    "path": "properties.kubernetesVersion",
                    "regex": "1.10.*"
                }
            },
            {
                "not": {
                    "path": "properties.kubernetesVersion",
                    "regex": "1.9.*"
                }
            },
            {
                "not": {
                    "path": "properties.kubernetesVersion",
                    "regex": "1.8.*"
                }
            },
            {
                "not": {
                    "path": "properties.kubernetesVersion",
                    "regex": "1.7.*"
                }
            },
            {
                "not": {
                    "path": "properties.kubernetesVersion",
                    "regex": "1.6.*"
                }
            },
            {
                "not": {
                    "path": "properties.kubernetesVersion",
                    "regex": "1.5.*"
                }
            },
            {
                "not": {
                    "path": "properties.kubernetesVersion",
                    "regex": "1.4.*"
                }
            },
            {
                "not": {
                    "path": "properties.kubernetesVersion",
                    "regex": "1.3.*"
                }
            },
            {
                "not": {
                    "path": "properties.kubernetesVersion",
                    "regex": "1.2.*"
                }
            },
            {
                "not": {
                    "path": "properties.kubernetesVersion",
                    "regex": "1.1.*"
                }
            },
            {
                "not": {
                    "path": "properties.kubernetesVersion",
                    "regex": "1.0.*"
                }
            }
        ]
    }
}
```

### LINKED ISSUE #90: Add support for the not operator
This operator is a logical not. It negates the evaluation in which it encompasses.

## Example
``` javascript
{
    "not": {
        "regex": "admin" // Evaluates to `false`
    } // Evaluates to `true` because it's the logical 'not' of the sub-evaluation that is `false`
}
```

Related to #44, #49

yane3628:
- Add not operator (Closes #90)
- Update tests for all operators
- Add tests for not operator
- Update documentation for not operator
- Add rule Kubernetes_Upgrade_Non-Vulnerable_Version and its tests(Closes #49)

### DIFF
--- a/src/Analyzer.Core/Rules/BuiltInRules.json
+++ b/src/Analyzer.Core/Rules/BuiltInRules.json
@@ -644,109 +644,13 @@
         {
           "not": {
             "path": "properties.kubernetesVersion",
-            "in": [
-              "1.13.4",
-              "1.13.3",
-              "1.13.2",
-              "1.13.1",
-              "1.13.0"
-            ]
+            "regex": "^1\\.((11\\.[0-8])|(12\\.[0-6])|(13\\.[0-4]))"
           }
         },
         {
           "not": {
             "path": "properties.kubernetesVersion",
-            "in": [
-              "1.12.6",
-              "1.12.5",
-              "1.12.4",
-              "1.12.3",
-              "1.12.2",
-              "1.12.1",
-              "1.12.0"
-            ]
-          }
-        },
-        {
-          "not": {
-            "path": "properties.kubernetesVersion",
-            "in": [
-              "1.11.8",
-              "1.11.7",
-              "1.11.6",
-              "1.11.5",
-              "1.11.4",
-              "1.11.3",
-              "1.11.2",
-              "1.11.1",
-              "1.11.0"
-            ]
-          }
-        },
-        {
-          "not": {
-            "path": "properties.kubernetesVersion",
-            "regex": "^1\\.10\\."
-          }
-        },
-        {
-          "not": {
-            "path": "properties.kubernetesVersion",
-            "regex": "^1\\.9\\."
-          }
-        },
-        {
-          "not": {
-            "path": "properties.kubernetesVersion",
-            "regex": "^1\\.8\\."
-          }
-        },
-        {
-          "not": {
-            "path": "properties.kubernetesVersion",
-            "regex": "^1\\.7\\."
-          }
-        },
-        {
-          "not": {
-            "path": "properties.kubernetesVersion",
-            "regex": "^1\\.6\\."
-          }
-        },
-        {
-          "not": {
-            "path": "properties.kubernetesVersion",
-            "regex": "^1\\.5\\."
-          }
-        },
-        {
-          "not": {
-            "path": "properties.kubernetesVersion",
-            "regex": "^1\\.4\\."
-          }
-        },
-        {
-          "not": {
-            "path": "properties.kubernetesVersion",
-            "regex": "^1\\.3\\."
-          }
-        },
-        {
-          "not": {
-            "path": "properties.kubernetesVersion",
-            "regex": "^1\\.2\\."
-          }
-        },
-        {
-          "not": {
-            "path": "properties.kubernetesVersion",
-            "regex": "^1\\.1\\."
-          }
-        },
-        {
-          "not": {
-            "path": "properties.kubernetesVersion",
-            "regex": "^1\\.0\\."
+            "regex": "^1\\.(([0-9]|10)\\.\\d+)"
           }
         }
       ]
--- a/src/Analyzer.Core/Rules/BuiltInRules.json
+++ b/src/Analyzer.Core/Rules/BuiltInRules.json
@@ -634,6 +634,125 @@
     }
   },
   {
+    "name": "Kubernetes_Upgrade_Non-Vulnerable_Version",
+    "description": "Kubernetes Services should be upgraded to a non-vulnerable Kubernetes version",
+    "recommendation": "Upgrade to a non-vulnerable Kubernetes version (1.11.9+, 1.12.7+, 1.13.5+, and 1.14.0+)",
+    "helpUri": "https://github.com/Azure/template-analyzer/blob/main/docs/built-in-bpa-rules.md/#kubernetes-services-should-be-upgraded-to-a-non-vulnerable-kubernetes-version",
+    "evaluation": {
+      "resourceType": "Microsoft.ContainerService/managedClusters",
+      "allOf": [
+        {
+          "not": {
+            "path": "properties.kubernetesVersion",
+            "in": [
+              "1.13.4",
+              "1.13.3",
+              "1.13.2",
+              "1.13.1",
+              "1.13.0"
+            ]
+          }
+        },
+        {
+          "not": {
+            "path": "properties.kubernetesVersion",
+            "in": [
+              "1.12.6",
+              "1.12.5",
+              "1.12.4",
+              "1.12.3",
+              "1.12.2",
+              "1.12.1",
+              "1.12.0"
+            ]
+          }
+        },
+        {
+          "not": {
+            "path": "properties.kubernetesVersion",
+            "in": [
+              "1.11.8",
+              "1.11.7",
+              "1.11.6",
+              "1.11.5",
+              "1.11.4",
+              "1.11.3",
+              "1.11.2",
+              "1.11.1",
+              "1.11.0"
+            ]
+          }
+        },
+        {
+          "not": {
+            "path": "properties.kubernetesVersion",
+            "regex": "1.10.*"
+          }
+        },
+        {
+          "not": {
+            "path": "properties.kubernetesVersion",
+            "regex": "1.9.*"
+          }
+        },
+        {
+          "not": {
+            "path": "properties.kubernetesVersion",
+            "regex": "1.8.*"
+          }
+        },
+        {
+          "not": {
+            "path": "properties.kubernetesVersion",
+            "regex": "1.7.*"
+          }
+        },
+        {
+          "not": {
+            "path": "properties.kubernetesVersion",
+            "regex": "1.6.*"
+          }
+        },
+        {
+          "not": {
+            "path": "properties.kubernetesVersion",
+            "regex": "1.5.*"
+          }
+        },
+        {
+          "not": {
+            "path": "properties.kubernetesVersion",
+            "regex": "1.4.*"
+          }
+        },
+        {
+          "not": {
+            "path": "properties.kubernetesVersion",
+            "regex": "1.3.*"
+          }
+        },
+        {
+          "not": {
+            "path": "properties.kubernetesVersion",
+            "regex": "1.2.*"
+          }
+        },
+        {
+          "not": {
+            "path": "properties.kubernetesVersion",
+            "regex": "1.1.*"
+          }
+        },
+        {
+          "not": {
+            "path": "properties.kubernetesVersion",
+            "regex": "1.0.*"
+          }
+        }
+      ]
+    }
+  },
+  {
     "name": "ServiceFabric_Client_AAD_Auth_Only",
     "description": "Service Fabric clusters should only use Azure Active Directory for client authentication",
     "recommendation": "Enable AAD client authentication on your Service Fabric clusters",
--- a/src/Analyzer.Core/Rules/BuiltInRules.json
+++ b/src/Analyzer.Core/Rules/BuiltInRules.json
@@ -686,67 +686,67 @@
         {
           "not": {
             "path": "properties.kubernetesVersion",
-            "regex": "1.10.*"
+            "regex": "^1\\.10\\."
           }
         },
         {
           "not": {
             "path": "properties.kubernetesVersion",
-            "regex": "1.9.*"
+            "regex": "^1\\.9\\."
           }
         },
         {
           "not": {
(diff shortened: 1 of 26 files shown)


src/Analyzer.JsonRuleEngine.UnitTests/InOperatorTests.cs
```diff
@@ -36,6 +36,32 @@ public void EvaluateExpression_ValidDataType_ReturnsExpectedEvaluationResult(boo
             Assert.AreEqual(evaluationResult, inOperator.EvaluateExpression(desiredValueJToken));
         }
 
+        [DataTestMethod]
```
VeraBE, line 39: Instead of duplicating the whole test, can't we also create an inOperator with `isNegative: true` in the original test function and do both validations there?

VeraBE: I think the same comment applies for other operators tests

src/Analyzer.JsonRuleEngine/Operators/ExistsOperator.cs
```diff
@@ -38,6 +38,6 @@ public ExistsOperator(bool specifiedValue, bool isNegative)
         /// <param name="tokenToEvaluate">The JToken to evaluate.</param>
         /// <returns>A value indicating whether or not the evaluation passed.</returns>
         public override bool EvaluateExpression(JToken tokenToEvaluate)
-            => (tokenToEvaluate != null) == EffectiveValue;
+            => (tokenToEvaluate != null) == EffectiveValue ^ this.IsNegative;
```
VeraBE, line 41: I'm not following why a XOR is a good choice for these operations

If EffectiveValue == true, and IsNegative == true, then EffectiveValue ^ this.IsNegative == true (I understand ^ has precendence over ==), but in that case wouldn't we want to check if (tokenToEvaluate != null) == false?

src/Analyzer.JsonRuleEngine/Schema/AllOfExpressionDefinition.cs
```diff
@@ -24,9 +24,20 @@ internal class AllOfExpressionDefinition : ExpressionDefinition
...
+        public override Expression ToExpression(ILineNumberResolver jsonLineNumberResolver, bool isNegative = false)
+        {
+            if (!isNegative)
+            {
+                return new AllOfExpression(this.AllOf.Select(e => e.ToExpression(jsonLineNumberResolver, isNegative)).ToArray(), GetCommonProperties(jsonLineNumberResolver));
+            }
+            else
+            {
+                // De Morgan's Law
+                return new AnyOfExpression(this.AllOf.Select(e => e.ToExpression(jsonLineNumberResolver, isNegative)).ToArray(), GetCommonProperties(jsonLineNumberResolver));
```
nonik0, line 38: For De Morgan's Law wouldn't the individual items then each be negated? !(A&B) === !A | !B

e.g. `return new AnyOfExpression(this.AllOf.Select(e =>e.ToExpression(jsonLineNumberResolver, !isNegative)...`

Tests will cover this so will be straightened out. :)

src/Analyzer.JsonRuleEngine/Schema/AnyOfExpressionDefinition.cs
```diff
@@ -24,9 +24,20 @@ internal class AnyOfExpressionDefinition : ExpressionDefinition
...
+        public override Expression ToExpression(ILineNumberResolver jsonLineNumberResolver, bool isNegative = false)
+        {
+            if (!isNegative)
+            {
+                return new AnyOfExpression(this.AnyOf.Select(e => e.ToExpression(jsonLineNumberResolver, isNegative)).ToArray(), GetCommonProperties(jsonLineNumberResolver));
+            }
+            else
+            {
+                // De Morgan's Law
+                return new AllOfExpression(this.AnyOf.Select(e => e.ToExpression(jsonLineNumberResolver, isNegative)).ToArray(), GetCommonProperties(jsonLineNumberResolver));
```
nonik0, line 38: Same as above.

src/Analyzer.JsonRuleEngine/Schema/AllOfExpressionDefinition.cs
yane3628: It's what Johnathon said. If it helps we can hardcode the values, or we can just add a comment saying that isNegative is true in this case.

src/Analyzer.JsonRuleEngine.UnitTests/AllOfExpressionDefinitionTests.cs
```diff
@@ -91,6 +129,42 @@ public void ToExpression_NestedAllOf_ReturnsArrayOfExpectedExpressions()
             Assert.IsInstanceOfType(allOfExpression.AllOf.Last(), typeof(LeafExpression));
         }
 
+        [TestMethod]
+        public void ToExpressionNegated_NestedAllOf_ReturnsArrayOfExpectedExpressions()
```
yane3628, line 133: Is it possible to collapse this test with the non-negated version? Same for the non-nested test and for the AnyOf tests

src/Analyzer.JsonRuleEngine.UnitTests/AllOfExpressionDefinitionTests.cs
```diff
@@ -55,6 +55,44 @@ public void ToExpression_ValidExpressions_ReturnsArrayOfExpectedExpressions(int
             }
```
yane3628, line 55: mockResolver can be readonly

yane3628 (commented):
@nonik0

yane3628:
We also need to update the documentation to indicate the not operator is available and add one built-in rule to make sure it works end to end.

yane3628:
We are missing the tests for the new built-in rule, but I'd like to wait to close https://github.com/Azure/template-analyzer/pull/169 first since both add tests for AKS.

nonik0:
> We also need to update the documentation to indicate the not operator is available and add one built-in rule to make sure it works end to end. @nonik0 I'll start doing these parts.

OK this was my one concern so sounds good.

src/Analyzer.JsonRuleEngine.UnitTests/AllOfExpressionDefinitionTests.cs
nonik0: Yeah it is possible and that's what I started out doing initially here, but I ultimately just split the tests out. Trying to have both scenarios run in one test would have been about the same number of lines as the two tests themselves, but a whole lot more convoluted. Since the negation causes the resultant expression object types to be different (AllOf/AnyOf), the rest of the code to verify the scenario gets less clean. Would involve lots of casting, using dynamic objects, or lots of if/else statements, etc.

src/Analyzer.JsonRuleEngine.UnitTests/AllOfExpressionDefinitionTests.cs
```diff
@@ -33,12 +33,12 @@ public void ToExpression_ValidExpressions_ReturnsArrayOfExpectedExpressions(int
             };
 
             // Act
-            var allOfExpression = allOfExpressionDefinition.ToExpression(mockResolver) as AllOfExpression;
+            var allOfExpression = allOfExpressionDefinition.ToExpression(mockResolver, false) as AllOfExpression;
```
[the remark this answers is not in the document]
yane3628: I think so. It's been a while but it's possible if it wasn't specified it would act strangely.

src/Analyzer.Core/Rules/BuiltInRules.json
```diff
@@ -633,6 +633,125 @@
...
+              "1.11.3",
+              "1.11.2",
+              "1.11.1",
+              "1.11.0"
+            ]
+          }
+        },
+        {
+          "not": {
+            "path": "properties.kubernetesVersion",
```
VeraBE, line 688: Can't we collapse all these last not properties with a regex like:
```
^1.[0-9].*
```
And maybe add another for the number 10, or:
```
^1.[0|1|2|3|4|5|6|7|8|9|10].*
```

src/Analyzer.Core/Rules/BuiltInRules.json
```diff
@@ -633,6 +633,125 @@
...
+  {
+    "name": "Kubernetes_Upgrade_Non-Vulnerable_Version",
+    "description": "Kubernetes Services should be upgraded to a non-vulnerable Kubernetes version",
+    "recommendation": "Upgrade to a non-vulnerable Kubernetes version (1.11.9+, 1.12.7+, 1.13.5+, and 1.14.0+)",
+    "helpUri": "https://github.com/Azure/template-analyzer/blob/main/docs/built-in-bpa-rules.md/#kubernetes-services-should-be-upgraded-to-a-non-vulnerable-kubernetes-version",
+    "evaluation": {
+      "resourceType": "Microsoft.ContainerService/managedClusters",
+      "allOf": [
+        {
+          "not": {
```
VeraBE, line 645: These could also be simplified with something like this for each:

```
{
  "not": {
    "path": "properties.kubernetesVersion",
    "regex": "^1.11.[0-8]*"
  }
}
```
Can't think of a way right now to avoid having three different regex for 11, 12 and 13

src/Analyzer.Core/Rules/BuiltInRules.json
VeraBE: Wait, in fact we can collapse all of them with:

```
{
  "not": {
    "path": "properties.kubernetesVersion",
    "regex": ""^1.((([0-9]|10).*)|(11.[0-8])|(12.[0-6])|(13.[0-4]))"
  }
}
```

yane3628: I was thinking about this, but for readability purposes I decided to leave it as is. I'm open to changing it though.

VeraBE: I like the regex better because for a programmer it shouldn't be that hard, it's just a couple of ranges and a couple of disjunctions, especially if you put it in an editor that shows you the relationship between the parenthesis and stuff like that (I used https://regex101.com/ to write it :)). And for a non-programmer, I feel that the rule description is clear enough that they shouldn't need to look at the code

JohnathonMohr: When I wrote a prototype for this a while back (during rule language creation) I did it in a single regex.  I do like the clarity and readability of separating it out though.  I'm ok with either method, but a hybrid might bring the best of both (clarity and compactness), like a regex for each minor version after 1.10, and one for everything <10?

yane3628: You're suggesting having two regex patterns, one for the ones that are listed in the list form and another for the ones that are currently represented as regex?

anaismiller (approved):

src/Analyzer.Core/Rules/BuiltInRules.json
nonik0: @yane3628 Based off how I read I think he meant four: for <=1.10, 1.11, 1.12, and 1.13. However, I would vote for the two like you said, which is just taking Vera's regex and basically splitting it in half:
^1.(([0-9]|10).*)
^1.((11.[0-8])|(12.[0-6])|(13.[0-4]))

As far as regexes go, these are fairly legible to me.

JohnathonMohr: Any of these sound good to me :)

yane3628: Updated it. I will also add that by having them all separated, it is much clearer what the tests should look like.

VeraBE (approved):

Issue #153, opened 2021-07-08
Add tests for built-in rules


VeraBE:
Now that we have a test framework for rules: https://github.com/Azure/template-analyzer/tree/main/src/Analyzer.Core.BuiltInRuleTests

PR #169, opened 2021-09-01
Add tests for the below built-in rules

yane3628:
- Authorized IP ranges should be defined on Kubernetes Services
- Automation account variables should be encrypted
- CORS should not allow every resource to access your API App
- CORS should not allow every resource to access your Function Apps
- CORS should not allow every resource to access your Web Applications

Working towards #153

### DIFF
--- a/src/Analyzer.Core.BuiltInRuleTests/TestTemplates/AppServices.badtemplate
+++ b/src/Analyzer.Core.BuiltInRuleTests/TestTemplates/AppServices.badtemplate
@@ -1,15 +1,15 @@
 {
-    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
-    "contentVersion": "1.0.0.0",
-    "parameters": {
-        "location": {
-            "type": "string",
-            "defaultValue": "[resourceGroup().location]",
-            "metadata": {
-                "description": "Location for all resources."
-            }
-        }
-    },
+  "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "location": {
+      "type": "string",
+      "defaultValue": "[resourceGroup().location]",
+      "metadata": {
+        "description": "Location for all resources."
+      }
+    }
+  },
   "resources": [
     {
       "apiVersion": "2019-08-01",
--- a/src/Analyzer.Core.BuiltInRuleTests/TestTemplates/AppServices.badtemplate
+++ b/src/Analyzer.Core.BuiltInRuleTests/TestTemplates/AppServices.badtemplate
@@ -145,11 +145,7 @@
       "kind": "api",
       "name": "ApiApp_RestrictedCORSAccess_EmbeddedSitesConfig",
       "location": "[parameters('location')]",
-      "dependsOn": [
-        "[resourceId('Microsoft.Web/serverfarms', 'serverFarm')]"
-      ],
       "properties": {
-        "serverFarmId": "[resourceId('Microsoft.Web/serverfarms', 'serverFarm')]",
         "httpsOnly": true,
         "siteConfig": {
           "cors": {
@@ -166,11 +162,7 @@
       "kind": "api",
       "name": "ApiApp_NoSitesConfig",
       "location": "[parameters('location')]",
-      "dependsOn": [
-        "[resourceId('Microsoft.Web/serverfarms', 'serverFarm')]"
-      ],
       "properties": {
-        "serverFarmId": "[resourceId('Microsoft.Web/serverfarms', 'serverFarm')]",
         "httpsOnly": true
       }
     },
@@ -185,7 +177,6 @@
         "FunctionApp_NoSitesConfig"
       ],
       "properties": {
-        "serverFarmId": "[resourceId('Microsoft.Web/serverfarms', 'serverFarm')]",
         "cors": {
           "allowedOrigins": [
             "someIP"
@@ -199,11 +190,7 @@
       "kind": "api",
       "name": "ApiApp_UnrestrictedCORSAccess_EmbeddedSitesConfig",
       "location": "[parameters('location')]",
-      "dependsOn": [
-        "[resourceId('Microsoft.Web/serverfarms', 'serverFarm')]"
-      ],
       "properties": {
-        "serverFarmId": "[resourceId('Microsoft.Web/serverfarms', 'serverFarm')]",
         "httpsOnly": true,
         "siteConfig": {
           "cors": {
@@ -226,7 +213,6 @@
         "FunctionApp_NoSitesConfig"
       ],
       "properties": {
-        "serverFarmId": "[resourceId('Microsoft.Web/serverfarms', 'serverFarm')]",
         "cors": {
           "allowedOrigins": [
             "*"
@@ -240,11 +226,7 @@
       "kind": "app",
       "name": "WebApp_RestrictedCORSAccess_EmbeddedSitesConfig",
       "location": "[parameters('location')]",
-      "dependsOn": [
-        "[resourceId('Microsoft.Web/serverfarms', 'serverFarm')]"
-      ],
       "properties": {
-        "serverFarmId": "[resourceId('Microsoft.Web/serverfarms', 'serverFarm')]",
         "httpsOnly": true,
         "siteConfig": {
           "cors": {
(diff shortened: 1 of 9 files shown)


src/Analyzer.Core.BuiltInRuleTests/TestTemplates/AppServices.badtemplate
```diff
@@ -10,134 +10,363 @@
...
+      "apiVersion": "2019-08-01",
+      "type": "Microsoft.Web/sites",
+      "kind": "api",
+      "name": "ApiApp_RestrictedCORSAccess_EmbeddedSitesConfig",
+      "location": "[parameters('location')]",
+      "dependsOn": [
+        "[resourceId('Microsoft.Web/serverfarms', 'serverFarm')]"
+      ],
+      "properties": {
+        "serverFarmId": "[resourceId('Microsoft.Web/serverfarms', 'serverFarm')]",
```
VeraBE, line 152: Why do you add a serverFarmId in every resource? I don't see any rule that uses that property

src/Analyzer.Core.BuiltInRuleTests/Tests/AppServiceApiApp_RestrictCORSAccess.json
```diff
@@ -0,0 +1,15 @@
+[
+    {
+        "Template": "AppServices",
+        "ReportedFailures": [
+            {
```
VeraBE, line 5: Same discussion as in Teams about rules with almost repeated code, I assume we will handle this situation later?

src/Analyzer.Core.BuiltInRuleTests/Tests/Kubernetes_Must_Restrict_Network_Access.json
```diff
@@ -0,0 +1,19 @@
+[
+  {
+    "Template": "Kubernetes",
+    "ReportedFailures": [
+      {
+        "LineNumber": 42,
+        "Description": "No Network restrictions defined"
+      },
+      {
+        "LineNumber": 62,
```
anaismiller, line 10: Why is 62 flagged, but not 51 or 73? If it because 73 has 75 and 51 is true on 52?

src/Analyzer.Core.BuiltInRuleTests/TestTemplates/AppServices.badtemplate
```diff
@@ -10,134 +10,363 @@
...
+              "*"
+            ]
+          }
+        }
+      }
+    },
+    {
+      "apiVersion": "2019-08-01",
+      "type": "Microsoft.Web/sites/config",
+      "name": "SitesConfig_UnrestrictedCORSAccess_web",
```
anaismiller, line 221: Just making sure that it is allowed to not have a "kind" property.

src/Analyzer.Core.BuiltInRuleTests/TestTemplates/AppServices.badtemplate
```diff
@@ -10,134 +10,363 @@
...
+              "someIP",
+              "*"
+            ]
+          }
+        }
+      }
+    },
+    {
+      "apiVersion": "2019-08-01",
+      "type": "Microsoft.Web/sites/config",
```
anaismiller, line 220: I don't see the "httpsOnly" property in this one. Is that an issue?

src/Analyzer.Core.BuiltInRuleTests/TestTemplates/AppServices.badtemplate
```diff
@@ -10,134 +10,363 @@
...
+      "name": "SitesConfig_UnrestrictedCORSAccess_web",
+      "location": "[parameters('location')]",
+      "dependsOn": [
+        "ApiApp_NoSitesConfig",
```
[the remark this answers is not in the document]
yane3628: This is to reuse the sitesconfig when testing the sitesconfig was declared separately from the sites. To reduce duplication of code.

src/Analyzer.Core.BuiltInRuleTests/TestTemplates/AppServices.badtemplate
yane3628: I can remove it. I just added it because the other app services from the HTTPS tests used it.

src/Analyzer.Core.BuiltInRuleTests/Tests/AppServiceApiApp_RestrictCORSAccess.json
yane3628: We can discuss options here or open an issue and decide later. Thoughts?

src/Analyzer.Core.BuiltInRuleTests/Tests/Kubernetes_Must_Restrict_Network_Access.json
yane3628: The rule requires at least one to be true. Since on line 52 enablePrivateCluster is set to true, this satisfies one condition of the OR statement.

src/Analyzer.Core.BuiltInRuleTests/TestTemplates/AppServices.badtemplate
yane3628: Kind is not required for resources of type sitesconfig

src/Analyzer.Core.BuiltInRuleTests/TestTemplates/AppServices.badtemplate
yane3628: httpsOnly is a property for resources of type sites not sites/config so it's not needed here.

src/Analyzer.Core.BuiltInRuleTests/Tests/AppServiceApiApp_RestrictCORSAccess.json
VeraBE: I like the second option :)

anaismiller (approved):

VeraBE (approved):

src/Analyzer.Core.BuiltInRuleTests/TestTemplates/AppServices.badtemplate
```diff
@@ -10,134 +10,321 @@
...
+      "name": "WebApp_HttpsTrue",
+      "location": "[parameters('location')]",
+      "dependsOn": [
+        "[resourceId('Microsoft.Web/serverfarms', 'serverFarm')]"
+      ],
+      "properties": {
+        "serverFarmId": "[resourceId('Microsoft.Web/serverfarms', 'serverFarm')]",
+        "httpsOnly": true
+      }
+    },
```
JohnathonMohr, line 141: Can you fix the indentation in this file?  Looks like the `resources` array isn't indented enough (resulting in the large diff)

src/Analyzer.Core.BuiltInRuleTests/TestTemplates/AppServices.badtemplate
```diff
@@ -10,134 +10,321 @@
...
+    {
+      "apiVersion": "2019-08-01",
+      "type": "Microsoft.Web/sites",
+      "kind": "functionapp",
+      "name": "FunctionApp_NoSitesConfig",
+      "location": "[parameters('location')]",
+      "properties": {
+        "httpsOnly": true
+      }
```
JohnathonMohr, line 327: Can you add a sites/config resource that isn't attached to a resource defined in the template (instead would update a resource already deployed) to verify it doesn't trigger a false positive?

JohnathonMohr (changes requested):

src/Analyzer.Core.BuiltInRuleTests/Tests/AppServiceApiApp_RestrictCORSAccess.json
yane3628: Make a new issue https://github.com/Azure/template-analyzer/issues/172

src/Analyzer.Core.BuiltInRuleTests/TestTemplates/AppServices.badtemplate
yane3628: Since all the other badtemplate files have less indentation, I'm going to reduce the indentation for lines 2-12 too match the others.

src/Analyzer.Core.BuiltInRuleTests/TestTemplates/AppServices.badtemplate
yane3628: If I add this, it would never trigger regardless of the CORS setting since the rule doesn't check stand-alone sites/configs. Can you elaborate on why you think we should add this?

JohnathonMohr: Yes, what you've described is the behavior it should validate - i.e. it shouldn't trigger (which would be a false positive if it did).  Without adding that resource, a regression that flags it incorrectly (returning the false positive) might not be caught.

JohnathonMohr: @yane3628 I take this back.  Looking at the change, it really wouldn't be a FP if BPA flagged it, we just don't have a great way to do it now.  This can be reverted to how you had it.

JohnathonMohr (approved):

PR #171, opened 2021-09-07
Add tests for some built-in rules

VeraBE:
- Diagnostic logs in App Services should be enabled
- Only secure connections to your Azure Cache for Redis should be enabled
- FTPS only should be required in your API App

Working towards #153

Decided on not adding these for now, in case we decide soon how to handle almost repeated rules (#172):
- FTPS only should be required in your Function App
- FTPS only should be required in your Web App (some test cases were added to cover its main difference with the other similar rules)

Also renaming resources in some existing test templates to use the [mandatory format for resource names: number of segments matching with the resource type](https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/error-invalid-template#solution-2---incorrect-segment-lengths)


yane3628:
@VeraBE  Is this PR ready to be reviewed?

src/Analyzer.Core.BuiltInRuleTests/TestRunner.cs
```diff
@@ -54,6 +54,11 @@ public void TestRule(TestConfiguration ruleExpectations)
                                     .ToList())
                     .ToHashSet();
 
+                failingLines.UnionWith(thisRuleEvaluation.Results
+                .Where(r => !r.Passed)
+                .Select(r => r.LineNumber)
+                .ToHashSet());
+
```
anaismiller, line 61: Should these be removed since they were in Yanelis' PR - merge conflicts?

src/Analyzer.Core.BuiltInRuleTests/Tests/AppService_EnableDiagLogs.json
```diff
@@ -0,0 +1,35 @@
+[
+    {
+        "Template": "AppServicesLogs", // didn't use the existing AppServices template because AppServiceWebApp_HTTPS would trigger for all the new resources, and AppService_EnableDiagLogs for all the already existing ones
```
anaismiller, line 3: I personally think it is better to mesh them together so that the existing test results are not touched, but the resources work with this additional rule. I also see that when there are 500 rules, separate may be better and easier to work with, but more an issue for the future.

src/Analyzer.Core.BuiltInRuleTests/Tests/AppService_EnableDiagLogs.json
```diff
@@ -0,0 +1,35 @@
...
+        "ReportedFailures": [
+            // The resources between lines 14 and 55 are not reported because their kind property includes "functionapp" or "linux"
```
anaismiller, line 5: These comments were very helpful for me who is new to this, but other tests don't have this. Should this be standardized throughout all the tests? Or should a new property be added to the test framework that allows for these comments (and again standardized?)

src/Analyzer.Core.BuiltInRuleTests/TestTemplates/AppServicesFTPS.badtemplate
```diff
@@ -0,0 +1,91 @@
...
+        },
+        {
+            "apiVersion": "2019-08-01",
+            "type": "Microsoft.Web/sites",
+            "kind": "api",
+            "name": "ftpsStateDisabled",
+            "location": "[parameters('location')]",
+            "properties": {
+                "siteConfig": {
+                    "ftpsState": "Disabled"
```
anaismiller, line 64: Should this line 64 be flagged too?

src/Analyzer.Core.BuiltInRuleTests/TestTemplates/AppServicesFTPS.badtemplate
```diff
@@ -0,0 +1,91 @@
...
+                }
+            }
+        },
+        {
+            "apiVersion": "2019-08-01",
+            "type": "Microsoft.Web/sites",
```
anaismiller, line 70: you forgot the "config" at the end

anaismiller: Should this be flagged in a test?

src/Analyzer.Core.BuiltInRuleTests/TestRunner.cs
VeraBE: If I remove it the tests don't run properly, I'll handle the merge after Yanelis merges hers to main

src/Analyzer.Core.BuiltInRuleTests/Tests/AppService_EnableDiagLogs.json
VeraBE: To use the same file I would need to add extra properties to all the resources that were created for other rules so this one doesn't fail, and then add my resources, or try to mix some test cases editing the existing resources, but it wouldn't work for most and it would tangle both rules in a confusing way

src/Analyzer.Core.BuiltInRuleTests/Tests/AppService_EnableDiagLogs.json
VeraBE: This PR made me like even more the idea we already discussed of having one template file per rule, which would make it super easy to document the passing test directly next to the resource

src/Analyzer.Core.BuiltInRuleTests/TestTemplates/AppServicesFTPS.badtemplate
VeraBE: The rule checks that the value of ftpsState is in [ "FtpsOnly", "Disabled" ]

src/Analyzer.Core.BuiltInRuleTests/TestTemplates/AppServicesFTPS.badtemplate
VeraBE: The rule checks for values of ftpsState when specified in a resource of type Microsoft.Web/sites/config, and in one of type Microsoft.Web/sites

src/Analyzer.Core.BuiltInRuleTests/Tests/AppService_EnableDiagLogs.json
JohnathonMohr: These can also just be added as JSON properties if you want.  `"Description"` isn't actually part of the schema, but it doesn't hurt anything to have it defined.

JohnathonMohr: I'm not quite sure what you mean by documenting tests "next to the resource".

VeraBE: I mean that if the templates were linked only to one rule we could add those description tags, or comments, next to each template resource that's meant to pass the analysis to explain why is part of the test

src/Analyzer.Core.BuiltInRuleTests/Tests/AppService_EnableDiagLogs.json
JohnathonMohr: The tests were designed so you could use multiple template files for testing.  Did you try that?  Could that make it easier?

JohnathonMohr: As more tests (and therefore test template files) are added, my hope was that it would make writing tests easier, since you can just find templates that already fail the rule and refer to those lines in the test instead of making all new template files.

VeraBE: I'm not sure how that would help in this case, I created a new template and didn't feel the need to use the resources in other templates

VeraBE: I didn't consider that, I'll see if it helps, but I doubt it because the problem is precisely that when I used already existing templates it flagged too many resources, that wouldn't make that much sense to report in the specific rule test file

yane3628: Right now we are using the name of the properties to describe the resource. If we start using the same resources for multiple rules we'll have to use some other method to describe the rule. Maybe we can use the tags property in the resource and add a new tag for each scenario we're testing? This method would shift line numbers for other rules pretty often though.

src/Analyzer.Core.BuiltInRuleTests/TestTemplates/AppServicesLogs.badtemplate
```diff
@@ -0,0 +1,130 @@
...
+    "parameters": {
+        "location": {
+            "type": "string",
+            "defaultValue": "[resourceGroup().location]",
+            "metadata": {
+                "description": "Location for all resources."
+            }
+        }
+    },
+    "resources": [
```
yane3628, line 13: Can you add one where some properties are true and some false?

src/Analyzer.Core.BuiltInRuleTests/TestTemplates/RedisCache.badtemplate
```diff
@@ -0,0 +1,41 @@
+{
+    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
```
yane3628, line 2: Tabbing for the .badtemplates are different than the previous .badtemplate files

yane3628 (changes requested):

src/Analyzer.Core.BuiltInRuleTests/TestTemplates/RedisCache.badtemplate
VeraBE: Ugh most test files also had inconsistent identations

src/Analyzer.Core.BuiltInRuleTests/TestTemplates/AppServicesLogs.badtemplate
VeraBE: It's kind of hard to find the middle point between thoroughly testing an operator and testing the rule code. With our current testing system, I feel it makes sense to fully test the behavior of an operator when we add the first rule that uses it, but it wouldn't make sense to do it for all the rules. Maybe is something worth discussing? Should I create an issue to add specific end-to-end operator tests? Or to come up with some criteria about this for rule testing?

JohnathonMohr: Can you clarify what you mean @VeraBE?  I think the intent of different combinations of true/false here would be to ensure the rule is using the right operators (not to test the functionality of the operators the rule uses).  For example, if the rule were to be updated later to use different operators to try and simplify it, but the simplification missed some template scenarios, these tests should be able to detect and prevent that.

src/Analyzer.Core.BuiltInRuleTests/TestTemplates/AppServicesFTPS.badtemplate
```diff
@@ -0,0 +1,91 @@
...
+    "parameters": {
+        "location": {
+            "type": "string",
+            "defaultValue": "[resourceGroup().location]",
+            "metadata": {
+                "description": "Location for all resources."
+            }
+        }
+    },
+    "resources": [
```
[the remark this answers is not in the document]
VeraBE: Pushed a test case and started a discussion in Teams about another I was trying to write

src/Analyzer.Core.BuiltInRuleTests/TestTemplates/AppServicesLogs.badtemplate
VeraBE: I kind of take that comment back, I was not considering the situation of updating the rule later to use different operators. But I meant that, for example, if a rule is the first one to use the AnyOf operator I can see us wanting to test all the combinations (all the elements false, one at a time true, and all the other combinations), but I don't see us doing that for every rule, that could potentially have 10 elements in the AnyOf array

src/Analyzer.Core.BuiltInRuleTests/Tests/AppService_EnableDiagLogs.json
JohnathonMohr: Sounds good to me.

yane3628 (approved):

src/Analyzer.Core.BuiltInRuleTests/Tests/AppService_EnableDiagLogs.json
anaismiller: Also good with me.

anaismiller (approved):

PR #183, opened 2021-10-05
Define rule IDs

VeraBE:


anaismiller (approved):
I did not cross-check that a rule's new ruleId remains consistent throughout the code. I'm assuming it is mapped properly.

src/Analyzer.Core/Rules/BuiltInRules.json
```diff
@@ -1,6 +1,6 @@
 [
   {
-    "name": "AppService_EnableDiagLogs",
+    "id": "TA-000001",
     "description": "Diagnostic logs in App Services should be enabled",
     "recommendation": "Enable diagnostic Logs in App Services",
     "helpUri": "https://github.com/Azure/template-analyzer/blob/main/docs/built-in-bpa-rules.md/#diagnostic-logs-in-app-services-should-be-enabled",
```
yane3628, line 6: I think the helpUris need to be updated too since the header changed :(

VeraBE: Oh you're right, good catch, luckily I think I can do it in one edit with some Sublime text edition magic

src/Analyzer.Types/IEvaluation.cs
```diff
@@ -12,9 +12,9 @@ namespace Microsoft.Azure.Templates.Analyzer.Types
     public interface IEvaluation
     {
         /// <summary>
-        /// Gets the name of the rule this evaluation is for.
+        /// Gets the ID of the rule this evaluation is for.
         /// </summary>
-        public string RuleName { get; }
+        public string RuleID { get; }
```
JohnathonMohr, line 17: I think `RuleId` and `Id` (lowercase 'd') would be a better style, and are more inline with code style guidelines.

src/Analyzer.PowerShellRuleEngine/PowerShellRuleEngine.cs
```diff
@@ -78,24 +74,34 @@ public IEnumerable<IEvaluation> EvaluateRules(string templateFilePath)
...
+                    var ruleDescription = executionResult.Name + ". " + uniqueError.Key;
+                    evaluations.Add(new PowerShellRuleEvaluation(ruleIDs[uniqueError.Key], ruleDescription, false, evaluationResults));
                 }
             }
 
             return evaluations;
         }
 
-        private void AddErrorToDictionary(dynamic error, ref Dictionary<string, SortedSet<int>> uniqueErrors)
+        private void AddErrorToDictionaries(dynamic error, ref Dictionary<string, SortedSet<int>> uniqueErrors, ref Dictionary<string, string> ruleIDs)
```
JohnathonMohr, line 85: Since these dictionaries are never reassigned to, they don't need to be passed as `ref`.

src/Analyzer.Core/Rules/BuiltInRules.json
```diff
@@ -112,7 +112,7 @@
     }
   },
   {
-    "name": "AppServiceApiApp_HTTPS",
+    "id": "TA-000004",
     "description": "API App should only be accessible over HTTPS",
     "recommendation": "Use HTTPS to ensure server/service authentication and protect data in transit from network layer eavesdropping attacks",
     "helpUri": "https://github.com/Azure/template-analyzer/blob/main/docs/built-in-bpa-rules.md#api-app-should-only-be-accessible-over-https",
```
yane3628, line 118: This help uri was missed to be updated

src/Analyzer.PowerShellRuleEngine/PowerShellRuleEngine.cs
```diff
@@ -78,24 +74,34 @@ public IEnumerable<IEvaluation> EvaluateRules(string templateFilePath)
...
                 {
-                    lineNumber = error.TargetObject.lineNumber;
+                    if (targetObject.Properties["lineNumber"] != null)
+                    {
+                        lineNumber = error.TargetObject.lineNumber;
+                    }
+
+                    if (targetObject.Properties["ruleId"] != null)
+                    {
+                        ruleId = "TA-" + error.TargetObject.ruleId;
```
yane3628, line 103: Can you add a test for this case?

PR #200, opened 2021-11-16
Define severity, rule suppression, rule selection

### LINKED ISSUE #132: Support rule suppression [#53468]
Users should be able to specify a set of rules to ignore during the analysis

### LINKED ISSUE #133: Support rule selection [#53468]
Users should be able to specify the set of rules to execute during the analysis

### LINKED ISSUE #177: Classification of rule severity (e.g., warning, informational, etc.) [#53468]

anaismiller:
Closes #177 
Closes #132 
Closes #133


docs/authoring-json-rules.md
```diff
@@ -12,6 +12,7 @@ Here are the fields that make up a rule definition.
     "description": "Brief description of what the rule is evaluating",
     "recommendation": "Guidance describing what should be done to fix the issue if a template violates the rule",
     "helpUri": "URI to find more detailed information about the rule and how to fix a template",
+    "severity" : "Int value 1 - 3, with 1 being high and 3 being low, designating the importance of the rule",
```
VeraBE, line 15:
```suggestion
    "severity" : "Integer value between 1 and 3, with 1 being high and 3 being low, designating the importance of the rule",
```

docs/authoring-json-rules.md
```diff
@@ -20,6 +21,7 @@ Here are the fields that make up a rule definition.
 - The `id` should look like `TA-NNNNNN`, with `NNNNNN` being the next unused number according to the [rule ids already defined](https://github.com/Azure/template-analyzer/blob/main/docs/built-in-bpa-rules.md).
 - The `recommendation` should provide clear but concise guidance on how to modify a template if the rule fails.  If some details are somewhat complex, or the rule takes a bit more to understand, add those details to a guide accessible at the URI in `helpUri`.
 - The `helpUri` is optional, but it is good practice to include.  For built-in rules, this will point to a guide in the GitHub repository.
+- The 'severity' is optional. If no severity is provided, it default evalutes to 2.
```
VeraBE, line 24:
```suggestion
- The `severity` is optional. If no severity is provided, it evaluates to 2 by default.
```

docs/customizing-evaluation-outputs.md
```diff
@@ -0,0 +1,54 @@
+# Template BPA Customizing Evaluation Outputs
+<a name="note"></a>***Note**: The ARM Template BPA is currently in development. All features that have yet to be implemented have been flagged with an asterisk [\*].*
+
+## Overview
+Template BPA customization is authored in JSON.  Custom evaluation settings are written in a separate configuration and used on tool executing. More details on how to run Template BPA with a configuration file can be found [here](https://github.com/Azure/template-analyzer#using-the-template-bpa).
```
VeraBE, line 5:
```suggestion
Template BPA customization is authored in JSON.  Custom evaluation settings are written in a separate configuration and used on tool execution. More details on how to run Template BPA with a configuration file can be found [here](https://github.com/Azure/template-analyzer#using-the-template-bpa). 
```

docs/customizing-evaluation-outputs.md
```diff
@@ -0,0 +1,54 @@
+# Template BPA Customizing Evaluation Outputs
+<a name="note"></a>***Note**: The ARM Template BPA is currently in development. All features that have yet to be implemented have been flagged with an asterisk [\*].*
+
+## Overview
+Template BPA customization is authored in JSON.  Custom evaluation settings are written in a separate configuration and used on tool executing. More details on how to run Template BPA with a configuration file can be found [here](https://github.com/Azure/template-analyzer#using-the-template-bpa). 
+
+## Template BPA Rule Object
+Here are the fields that make up a custom configuration file.
```
VeraBE, line 8:
```suggestion
Here are the fields that make up a custom configuration file:
```

docs/customizing-evaluation-outputs.md
```diff
@@ -0,0 +1,54 @@
...
+```javascript
+{
+    "exclusions": { � }, // Parameters that are excluded from the BPA execution.  More details below.
```
VeraBE, line 11:
```suggestion
    "exclusions": { � }, // Parameters that are excluded from the BPA execution. More details below.
```

VeraBE: Same in the lines below

docs/customizing-evaluation-outputs.md
```diff
@@ -0,0 +1,54 @@
...
+    "inclusions": { � }, // Only parameters that will be included in the BPA execution.  More details below.
+    "output": { � }, // How the BPA execution output will be displayed to the user.  More details below.
+    "severityReMapping": { � } // Tuple of ruleId and new severity value.  Can change a rule's severity.
```
VeraBE, line 14:
```suggestion
    "severityReMapping": { � } // Tuple of rule id and new severity value. Can change a rule's severity.
```

docs/customizing-evaluation-outputs.md
```diff
@@ -0,0 +1,54 @@
...
+}
+```
+
+### Guidelines for exclusions object
+The `exclusions` is compromised of the following optional properties.
```
VeraBE, line 19:
```suggestion
The `exclusions` are comprised of the following optional properties:
```

docs/customizing-evaluation-outputs.md
```diff
@@ -0,0 +1,54 @@
...
+```javascript
+"exclusions": {
+    "severity": [int], // Means don't include rule with servities that are in this list in the BPA execution
```
VeraBE, line 22:
```suggestion
    "severity": [int], // Means don't include rules with severities that are in this list in the BPA execution
```

docs/customizing-evaluation-outputs.md
VeraBE: Same for the below sections

docs/customizing-evaluation-outputs.md
```diff
@@ -0,0 +1,54 @@
...
+### Guidelines for inclusions object
+The `inclusions` is compromised of the following optional properties.
+```javascript
+"inclusions": {
+    "ruleIds": ["ruleId"] // Means only include rules that are in this list in the BPA execution
+}
+```
+If the inclusions object contains values, then the exclusions object will be ignored. 
+
+### Guidelines for output object [[*]](#note)
```
VeraBE, line 37: Maybe it makes sense to merge this PR after the SARIF one? We will support file output in SARIF format and console output

docs/customizing-evaluation-outputs.md
```diff
@@ -0,0 +1,54 @@
...
+The `output` is compromised of the following optional properties.
+```javascript
+"output" : {
+    "file": boolean, 
+    "sortBy": ["higherSev", "alphabetic"]
+}
+```
+
+### Guidelines for severityReMapping tuple
+The `severityReMapping` is composed of a tuple of ruleId and new severity value. Below is an example of changing the severities on two rules.
```
VeraBE, line 48:
```suggestion
The `severityReMapping` is composed of a tuple of a rule id and a new severity value. Below is an example of changing the severities on two rules:
```

docs/customizing-evaluation-outputs.md
VeraBE, line 11: This work would also close #132 and #133, it probably makes sense to rename the PR

docs/customizing-evaluation-outputs.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Template BPA Rule Object
+Here are the fields that make up a custom configuration file.
+```javascript
+{
+    "exclusions": { � }, // Parameters that are excluded from the BPA execution.  More details below.
+    "inclusions": { � }, // Only parameters that will be included in the BPA execution.  More details below.
+    "output": { � }, // How the BPA execution output will be displayed to the user.  More details below.
+    "severityReMapping": { � } // Tuple of ruleId and new severity value.  Can change a rule's severity.
+}
```
VeraBE, line 15: It would make sense to me to support all the BPA parameters inside this file, maybe we could add the file paths here too? Or remove the options that are supported without this file, like the type of output

src/Analyzer.Core/TemplateAnalyzer.cs
```diff
@@ -39,7 +39,7 @@ public static TemplateAnalyzer Create()
             string rules;
             try
             {
-                rules = LoadRules();
+                rules = LoadRules();
```
VeraBE, line 42:
```suggestion
                rules = LoadRules();
```

src/Analyzer.JsonRuleEngine/Schemas/ConfigurationsDefinition.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) Microsoft Corporation.
```
VeraBE, line 1: Was the folder renamed intentionally? Just double checking

src/Analyzer.JsonRuleEngine/Schemas/ConfigurationsDefinition.cs
```diff
@@ -0,0 +1,31 @@
...
+        public ExclusionsConfigurationsDefinition ExclusionsConfigurationsDefinition { get; set; }
+
+        /// <summary>
+        /// Gets or sets the inclusion details of the configuration.
+        /// </summary>
+        [JsonProperty(PropertyName = "inclusions")]
+        public InclusionsConfigurationsDefinition InclusionsConfigurationsDefinition { get; set; }
+
+        /// <summary>
+        /// Gets or sets the severityReMapping property
```
VeraBE, line 27:
```suggestion
        /// Gets or sets the severity remapping of the configuration.
```

src/Analyzer.JsonRuleEngine/Schemas/ConfigurationsDefinition.cs
```diff
@@ -0,0 +1,31 @@
...
+        /// </summary>
+        public String severityReMapping { get; set; }
```
VeraBE, line 29: Refactor to uppercase like the other public variables?

src/Analyzer.JsonRuleEngine/Schemas/ConfigurationsDefinition.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Newtonsoft.Json;
```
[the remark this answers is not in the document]
VeraBE: Same in other files

src/Analyzer.JsonRuleEngine.UnitTests/ConfigurationsTests.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Templates.Analyzer.RuleEngines.JsonEngine.UnitTests
+{
+    [TestClass]
+    public class ConfigurationsTests
+    {
+        // Not sure what to put
```
VeraBE, line 13: Maybe you can mock the configuration objects to run the JSON engine with that and validate the config features? Some end to end with test config files would be ideal too

src/Analyzer.JsonRuleEngine/Schemas/ConfigurationsDefinition.cs
anaismiller: Yes. When I created a class, it auto made the namespace be `.Schema` , but all the "Schemas" have namespace `.Schemas`

docs/customizing-evaluation-outputs.md
VeraBE: I think we need to write "Closes #issue" in its own line for each so that the issues get closed when the PR is merged

JohnathonMohr: Right - when GitHub detects the "closing language", the issue will show up on the right side bar under "Linked issues"

docs/customizing-evaluation-outputs.md
anaismiller: Are we thinking we want to be able to define output both here in the configurations file and on the console, or do we only want on the console?

VeraBE: I like the idea of being consistent for all parameters (everything available in here too, or only the options that are not already available as console options) like I tried to say in another comment

src/Analyzer.Core/TemplateAnalyzer.cs
```diff
@@ -108,5 +113,32 @@ private static string LoadRules()
...
+        /// <param name="configurations">Optional Command-line Configurations paramater.</param>
+        /// <returns>Configurations file path if exists.</returns>
+        private string LoadConfiguration(string configurations)
+        {
+            try
+            {
+                if (configurations == null)
+                {
+                    var defaultPath = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location) +
+                        "/Configurations/Configuration.json";
```
VeraBE, line 129: It might be necessary to use something like https://docs.microsoft.com/en-us/dotnet/api/system.io.path.directoryseparatorchar?view=net-6.0 here, for the Linux tests to pass (two are failing now)

docs/authoring-json-rules.md
VeraBE: Unresolved as it is not fully fixed

docs/customizing-evaluation-outputs.md
VeraBE: Unresolved as it is not fully fixed

src/Analyzer.JsonRuleEngine/JsonRuleEngine.cs
```diff
@@ -49,6 +49,51 @@ public static JsonRuleEngine Create(string rawRuleDefinitions, BuildILineNumberR
             return new JsonRuleEngine(ParseRuleDefinitions(rawRuleDefinitions), jsonLineNumberResolverBuilder);
         }
 
+        /// <summary>
+        /// Modifies the rules to run based on values defined in the configurations file.
+        /// </summary>
+        /// <param name="configuration">The template configurations to filter and analyze.</param>
+        public void FilterRules(string configuration)
+        {
+            var result = new List<RuleDefinition>();
```
yane3628, line 58: I'd recommend renaming this variable. These aren't results (violations found in templates), but the rules themselves.

src/Analyzer.Core/TemplateAnalyzer.cs
```diff
@@ -85,6 +86,10 @@ public IEnumerable<IEvaluation> AnalyzeTemplate(string template, string paramete
 
             try
             {
+                var configuration = LoadConfiguration(configurations);
+                if (configuration != null)
+                    jsonRuleEngine.FilterRules(configuration);
```
yane3628, line 91: I think the rule filtering should be done at an earlier stage in the process. At this point, it would be filtering the rules at each evaluation of a template. It should likely instead happen in the TemplateAnalyzer.Create() or LoadRules().

yane3628: Create is called on line 31 in CommandLineParser.cs

src/Analyzer.Core/TemplateAnalyzer.cs
```diff
@@ -108,5 +113,32 @@ private static string LoadRules()
...
+        /// <summary>
+        /// Loads the configurations file from default directory if no file was passed as an optional parameter.
+        /// </summary>
+        /// <param name="configurations">Optional Command-line Configurations paramater.</param>
+        /// <returns>Configurations file path if exists.</returns>
+        private string LoadConfiguration(string configurations)
+        {
+            try
+            {
+                if (configurations == null)
```
yane3628, line 126: This is expecting a file to be present by default. Is this expected?

src/Analyzer.JsonRuleEngine/JsonRuleEngine.cs
```diff
@@ -49,6 +49,51 @@ public static JsonRuleEngine Create(string rawRuleDefinitions, BuildILineNumberR
             return new JsonRuleEngine(ParseRuleDefinitions(rawRuleDefinitions), jsonLineNumberResolverBuilder);
         }
 
+        /// <summary>
+        /// Modifies the rules to run based on values defined in the configurations file.
+        /// </summary>
+        /// <param name="configuration">The template configurations to filter and analyze.</param>
+        public void FilterRules(string configuration)
```
yane3628, line 56: This function needs tests and should be moved outside of the context of the rule engines. It should be somewhere in the TemplateAnalyzer code since the configurations should cover all rules across the engines.

src/Analyzer.JsonRuleEngine/Schemas/ConfigurationsDefinition.cs
yane3628: I think the configurations stuff should be at Analyzer.Core project not in the Rule Engine since it's scoped to the entire tool not to a particular engine in the tool. Do we need to make interfaces for these types of things to have consistency across our engines?

yane3628 (changes requested):

src/Analyzer.JsonRuleEngine/JsonRuleEngine.cs
VeraBE: Oh, I forgot about TTK, if we want this configuration file to work with those rules we also need to go through those rules

src/Analyzer.Reports/IReportWriter.cs
```diff
@@ -18,6 +18,7 @@ public interface IReportWriter : IDisposable
         /// <param name="evaluations">Evaluation list</param>
         /// <param name="templateFile">Template file to be analyzed</param>
         /// <param name="parametersFile">The parameter file to use when parsing the specified ARM template.</param>
-        void WriteResults(IEnumerable<Types.IEvaluation> evaluations, IFileInfo templateFile, IFileInfo parametersFile = null);
+        /// <param name="configurationsFile">The configurations file to use when running the analyzer</param>
+        void WriteResults(IEnumerable<Types.IEvaluation> evaluations, IFileInfo templateFile, IFileInfo parametersFile = null, IFileInfo configurationsFile = null);
```
JohnathonMohr, line 22: Since the configuration file applies to the entire run, it's better to pass that information in the constructor of a report writer instead of when writing results, since the config will apply globally and not just to a set of results.

src/Analyzer.JsonRuleEngine/Schemas/RuleDefinition.cs
```diff
@@ -35,6 +35,12 @@ internal class RuleDefinition
         [JsonProperty]
         public string HelpUri { get; set; }
 
+        /// <summary>
+        /// Gets or sets the severity of the rule.
+        /// </summary>
+        [JsonProperty]
+        public int Severity { get; set; }
```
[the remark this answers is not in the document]
JohnathonMohr: I also suggest making it nullable, so it's clearer when the severity is not defined in a rule.

src/Analyzer.JsonRuleEngine/JsonRuleEngine.cs
```diff
@@ -23,7 +23,7 @@ public class JsonRuleEngine : IRuleEngine
         public delegate ILineNumberResolver BuildILineNumberResolver(TemplateContext context);
 
         private readonly BuildILineNumberResolver BuildLineNumberResolver;
-        internal readonly IReadOnlyList<RuleDefinition> RuleDefinitions;
+        internal IList<RuleDefinition> RuleDefinitions;
```
JohnathonMohr, line 26: Keep the type `IReadOnlyList` - you can still assign a new value to it, but it will prevent changes to *List* itself.

JohnathonMohr (changes requested):

src/Analyzer.Core/TemplateAnalyzer.cs
JohnathonMohr: Create would be hard, since it's called before the command line is parsed.  But calling LoadRules before analysis would work instead.

src/Analyzer.JsonRuleEngine/Schemas/ConfigurationsDefinition.cs
JohnathonMohr: I don't know if we need an interface or not, but the config should probably be moved to Analyzer.Types.

src/Analyzer.JsonRuleEngine/JsonRuleEngine.cs
VeraBE: Here's the info on how to suppress/select TTK rules: https://github.com/Azure/arm-ttk/tree/master/arm-ttk The examples show one rule at a time but I think multiple are possible

docs/customizing-evaluation-outputs.md
anaismiller: fixed, thanks!

docs/customizing-evaluation-outputs.md
anaismiller: Removed `"output"`

README.md
```diff
@@ -33,6 +35,7 @@ Argument | Description
 **(Optional)** `-p` or `--parameters-file-path` | A [parameters file](https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/parameter-files)
 **(Optional)** `--report-format` | <p>*Console*: output results to the console in plain text. **(default)**</p>*Sarif*: output results to a file in [SARIF](https://sarifweb.azurewebsites.net) format.
 `-o` or `--output-file-path` | **(Required if `--report-format` is *Sarif*)**  File path to output SARIF results to.
+**(Optional)** `-c` or `--config-file-path` | A custom [configuration file](https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/parameter-files) with custom settings
```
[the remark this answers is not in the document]
anaismiller: good catch!

docs/authoring-json-rules.md
```diff
@@ -20,6 +21,7 @@ Here are the fields that make up a rule definition.
 - The `id` should look like `TA-NNNNNN`, with `NNNNNN` being the next unused number according to the [rule ids already defined](https://github.com/Azure/template-analyzer/blob/main/docs/built-in-bpa-rules.md).
 - The `recommendation` should provide clear but concise guidance on how to modify a template if the rule fails.  If some details are somewhat complex, or the rule takes a bit more to understand, add those details to a guide accessible at the URI in `helpUri`.
 - The `helpUri` is optional, but it is good practice to include.  For built-in rules, this will point to a guide in the GitHub repository.
+- The `severity` is optional. If no severity is provided, it evalutes to 2 by default.
```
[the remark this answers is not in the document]
anaismiller: @VeraBE is that ok with you? 
I had originally put "it default evaluates to 2."

docs/customizing-evaluation-outputs.md
```diff
@@ -0,0 +1,55 @@
...
+    "output": { � }, // How the BPA execution output will be displayed to the user. More details below.
+    "severityReMapping": { � } // Tuple of Id and new severity value. Can change a rule's severity.
+}
+```
+
+### Guidelines for exclusions object
+The `exclusions` object is compromised of the following optional properties:
+```javascript
+"exclusions": {
+    "severity": [int], // Means don't include rules with severities that are in this list in the BPA execution
```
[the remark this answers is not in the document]
anaismiller: Using "omitted" instead of "excluded"

docs/customizing-evaluation-outputs.md
```diff
@@ -0,0 +1,55 @@
...
+The `inclusions` object is compromised of the following optional properties:
+```javascript
+"inclusions": {
+    "ids": ["Id"] // Means only include rules that are in this list in the BPA execution
+}
+```
+If the inclusions object contains values, then the exclusions object will be ignored. 
+
+### Guidelines for output object [[*]](#note)
+**NOTE: `Output` is not yet supported.
```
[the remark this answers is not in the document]
anaismiller: I copied the format of `Evaluate` in the docs/authoring-json-rules.md

docs/customizing-evaluation-outputs.md
```diff
@@ -0,0 +1,55 @@
...
+Template BPA customization is authored in JSON.  Custom evaluation settings are written in a separate configuration and used on tool execution. More details on how to run Template BPA with a configuration file can be found [here](https://github.com/Azure/template-analyzer#using-the-template-bpa). 
+
+## Template BPA Rule Object
+Here are the fields that make up a custom configuration file:
+```javascript
+{
+    "exclusions": { � }, // Parameters that are excluded from the BPA execution. More details below.
+    "inclusions": { � }, // Only parameters that will be included in the BPA execution. More details below.
+    "output": { � }, // How the BPA execution output will be displayed to the user. More details below.
+    "severityReMapping": { � } // Tuple of Id and new severity value. Can change a rule's severity.
```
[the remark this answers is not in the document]
anaismiller: I like `severityOverides`. Updating to that name

src/Analyzer.Core/TemplateAnalyzer.cs
anaismiller: The intent was if a configurations param was not passed in w/ the cli, then see if there is one in the defaultPath to use. Does that make more sense?

docs/customizing-evaluation-outputs.md
```diff
@@ -0,0 +1,55 @@
...
+}
+```
+If the inclusions object contains values, then the exclusions object will be ignored. 
+
+### Guidelines for inclusions object
+The `inclusions` object is compromised of the following optional properties:
+```javascript
+"inclusions": {
+    "ids": ["Id"] // Means only include rules that are in this list in the BPA execution
+}
```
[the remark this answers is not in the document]
anaismiller: It can be. At the time, it seemed redundant to have it be in both as there are only three severity options. Should I add it in?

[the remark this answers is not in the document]
JohnathonMohr: Ah, you mean if someone wants to only include '3', they can just exclude '1' and '2'?  That could work.  If it's easy enough though, it might be more user-friendly to just support it in both ways.

src/Analyzer.JsonRuleEngine/Schemas/RuleDefinition.cs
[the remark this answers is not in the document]
anaismiller: `public enum Severity { high = 1, medium = 2, low = 3};` 
Something like this^? Where would I set this and I'm assuming the above lines changes to something like this? : 
`[JsonProperty]
        public Severity _Severity { get; set; }`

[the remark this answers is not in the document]
JohnathonMohr: This would be defined in the Types library.  And the property name can remain `Severity` - the compiler will understand the property name is different than the type.

src/Analyzer.Core/TemplateAnalyzer.cs
anaismiller: +1 for `Create()` = hard and that is true that it is doing it at each evaluation. Johnathon, what do you mean by "But calling LoadRules before analysis would work instead"?
Would calling it in CommandLineParse.cs in `SetupCommandLineAPI()` before the first call to `AnalyzeTemplate` and (re)moving all the `configurationsFilePath` references/actions?

JohnathonMohr: Oops, I meant `FilterRules` before analysis, not LoadRules.  Call if from the CLI in `AnalyzeDirectory` before analyzing any templates.  (You'd also need to make sure it's only done in `AnalyzeTemplate` if it isn't done for the directory already.)

Also opened #207 to cleanup growing commonality between template and directory analysis.

src/Analyzer.JsonRuleEngine/JsonRuleEngine.cs
anaismiller: Vera for TTK, because it is a cli argument, should this still be in here? / where does the loading of ttk rules happen in BPA?
Yanelis, Johnathon, I am confused by the decision. Are we agreeing that each rule engine will have their own FilterRules or it will be universal? This may get clearer based on where I move calling this function too.

JohnathonMohr: Others can chime in, but here's my opinion:
- An engine shouldn't run rules that aren't going to be reported, so the engine should filter the rules before analysis.
  - For this reason, I think filtering should stay in the context of the engine, since Analyzer.Core won't know how to filter.
- The way TTK runs currently is temporary, so I think we shouldn't put effort into filtering those results.
  - Instead, we should put our effort into creating the future PowerShell engine that TTK rules can be rewritten in.

src/Analyzer.Core/TemplateAnalyzer.cs
JohnathonMohr: Sorry, let me clarify again after regaining the context - likely what should happen is some new function like `LoadConfiguration` in Analyzer.Core should be invoked before any analysis.  That function would in turn call `FilterRules` on the rule engine.  After this, the templates are analyzed.

src/Analyzer.Core/TemplateAnalyzer.cs
JohnathonMohr: @anaismiller, I'm guessing the idea here is to always look to see if a user has defined a config at that path, and if they have, to go ahead and use it even if it wasn't specified, is that correct?  I'm guessing this because that file isn't part of this PR anywhere, so the user must define it on their own.

If the above is correct, this behavior needs to be documented, and I don't see that in the PR anywhere.  Can you please add that?  It should also at least be mentioned in the main README, with a link to the docs.

src/Analyzer.JsonRuleEngine/JsonRuleEngine.cs
```diff
@@ -49,6 +49,57 @@ public static JsonRuleEngine Create(string rawRuleDefinitions, BuildILineNumberR
             return new JsonRuleEngine(ParseRuleDefinitions(rawRuleDefinitions), jsonLineNumberResolverBuilder);
         }
 
+        /// <summary>
+        /// Modifies the rules to run based on values defined in the configurations file.
+        /// </summary>
+        /// <param name="configuration">The template configurations to filter and analyze.</param>
```
JohnathonMohr, line 55: Minor: this can be reworded so it doesn't sound like the configuration is being filtered.  Something like "The configuration specifying rule modifications."

src/Analyzer.JsonRuleEngine/JsonRuleEngine.cs
```diff
@@ -49,6 +49,57 @@ public static JsonRuleEngine Create(string rawRuleDefinitions, BuildILineNumberR
...
+            {
+                var includeSeverities = contents.InclusionsConfigurationDefinition.Severity;
+                var includeIds = contents.InclusionsConfigurationDefinition.Ids;
+                foreach (RuleDefinition rule in RuleDefinitions)
+                {
+                    if (includeSeverities.Contains(rule.Severity) || includeIds.Contains(rule.Id))
+                    {
+                        filteredRules.Add(rule);
+                    }
+                }
```
JohnathonMohr, line 80: Suggestion (I think this will work as-is, might need slight modification).  Could also do something similar in `else` block.
``` suggestion
                RuleDefinitions = RuleDefinitions.Where(r => includeSeverities.Contains(r.Severity) || includeIds.Contains(r.Id)).ToList().AsReadOnly();
```

src/Analyzer.JsonRuleEngine/JsonRuleEngine.cs
```diff
@@ -49,6 +49,57 @@ public static JsonRuleEngine Create(string rawRuleDefinitions, BuildILineNumberR
...
+                throw new JsonRuleEngineException($"Failed to parse configurations file.", e);
+            }
+
+            if (contents.InclusionsConfigurationDefinition != null)
+            {
+                var includeSeverities = contents.InclusionsConfigurationDefinition.Severity;
+                var includeIds = contents.InclusionsConfigurationDefinition.Ids;
+                foreach (RuleDefinition rule in RuleDefinitions)
+                {
+                    if (includeSeverities.Contains(rule.Severity) || includeIds.Contains(rule.Id))
```
JohnathonMohr, line 76: Just want to verify the logic here.  If the provided config file has:
```
"inclusions": {
    "severity": [ 1 ]
    "ids": [ "SomeRuleId" ]
}
```
... the result will be running all rules with severity 1 *and* 'SomeRuleId', even if that rule has a different severity than 1.  Is this what we want?  Should we make this behavior explicit in documentation?

src/Analyzer.JsonRuleEngine/JsonRuleEngine.cs
```diff
@@ -49,6 +49,57 @@ public static JsonRuleEngine Create(string rawRuleDefinitions, BuildILineNumberR
...
+                }
+            }
+            else
+            {
+                // Return original rule set - Configurations File does not have (supported) filters
+                return;
+            }
+            
+            // Update rules to select desired rules
+            RuleDefinitions = filteredRules;
```
JohnathonMohr, line 100: If you reassign this in the `if` statements, you won't need an extra `else` just to return early.

src/Analyzer.Types/ConfigurationDefinition.cs
```diff
@@ -0,0 +1,31 @@
...
+        /// <summary>
+        /// Gets or sets the inclusion details of the configuration.
+        /// </summary>
+        [JsonProperty(PropertyName = "inclusions")]
+        public InclusionsConfigurationDefinition InclusionsConfigurationDefinition { get; set; }
+
+        /// <summary>
+        /// Gets or sets the severity overrides of the configuration
+        /// </summary>
+        public String SeverityOverrides { get; set; }
```
JohnathonMohr, line 29: Since this isn't supported yet, don't add it to the schema.

src/Analyzer.Types/ExclusionsConfigurationDefinition.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Collections.Generic;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.Templates.Analyzer.Types
+{
+    /// <summary>
+    /// Gets or sets the ExclusionsConfigurationDefinition.
```
JohnathonMohr, line 10: Update summary (a class can't get or set something).  Something like `The exclusions defined in a <see cref="ConfigurationDefinition"/>`.

src/Analyzer.Types/InclusionsConfigurationDefinition.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Collections.Generic;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.Templates.Analyzer.Types
+{
+    /// <summary>
+    /// Gets or sets the InclusionsConfigurationDefinition.
```
JohnathonMohr, line 10: Update (similar to Exclusions)

src/Analyzer.JsonRuleEngine/JsonRuleEngine.cs
```diff
@@ -96,6 +147,7 @@ private static List<RuleDefinition> ParseRuleDefinitions(string rawRuleDefinitio
                 {
                     currentRule = rule.Id;
                     rule.Expression = rule.ExpressionDefinition.ToExpression();
+                    rule.Severity = rule.Severity == default ? 2 : rule.Severity;
```
JohnathonMohr, line 150: Instead of assigning a default value here, do it in the schema directly (see comment in RuleDefinition.cs).

src/Analyzer.JsonRuleEngine/Schemas/RuleDefinition.cs
JohnathonMohr, line 42: Set the default value here explicitly.
```cs
        public int Severity { get; set; } = 2;
```
Or when moving to enums:
```cs
        public Severity Severity { get; set; } = Severity.Medium;
```

JohnathonMohr (changes requested):

docs/customizing-evaluation-outputs.md
```diff
@@ -0,0 +1,45 @@
...
+```
+If the inclusions object contains values, then the exclusions object will be ignored. 
+
+### Inclusions Object
+The `inclusions` object is comprised of the following optional properties:
+```javascript
+"inclusions": {
+    "severity": [int], // List of severities to include in results. Any rules with matching severities will be included in results
+    "ids": ["Id"] // List of ids to include in results. Any rules with matching ids will be included in results
```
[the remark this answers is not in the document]
anaismiller: I moved it to the paragraph below instead w/ word "only" as it is both ids and severities together.

docs/customizing-evaluation-outputs.md
```diff
@@ -0,0 +1,45 @@
...
+}
```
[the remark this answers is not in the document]
anaismiller: I did not know that is how it worked.

src/Analyzer.Core/TemplateAnalyzer.cs
```diff
@@ -108,5 +113,32 @@ private static string LoadRules()
                     Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location),
                     "Rules/BuiltInRules.json"));
         }
+
+        /// <summary>
+        /// Loads the configurations file from default directory if no file was passed as an optional parameter.
+        /// </summary>
+        /// <param name="configurations">Optional Command-line Configurations paramater.</param>
+        /// <returns>Configurations file path if exists.</returns>
+        public string LoadConfiguration(string configurations)
```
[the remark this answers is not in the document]
anaismiller: Should the `LoadRules` function also be renamed?

src/Analyzer.JsonRuleEngine/JsonRuleEngine.cs
anaismiller: Yes, that is what we want. I believe I have done that in this latest update.

src/Analyzer.JsonRuleEngine/Schemas/RuleDefinition.cs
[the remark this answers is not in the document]
anaismiller: I am still unable to get it to work. 
In Types.IEvaluation:
`        /// <summary>
        /// The severity of the rule.
        /// </summary>
        public enum Severity {
            /// <summary>
            /// Severity High = 1
            /// </summary>
            High = 1,
            /// <summary>
            /// Severity Medium = 2
            /// </summary>
            Medium = 2,
            /// <summary>
            /// Severity Low = 3
            /// </summary>
            Low = 3
        };`

In JsonRuleEngine.RuleDefiniton
`        /// <summary>
        /// Gets or sets the severity of the rule.
        /// </summary>
        [JsonProperty]
        public Types.IEvaluation.Severity Severity { get; set; } = Types.IEvaluation.Severity.Medium;`

but, then all the other files complain and I don't have a getter for the enum.

src/Analyzer.JsonRuleEngine/JsonRuleEngine.cs
anaismiller: @yane3628 In the last discussion, you wanted me to move filtering into Analyzer.Core Template Analyzer from JsonRuleEngine without the addition of the interface. However, doing that means having to re-create/reference the RuleDefinition Schema in the JsonRuleEngine as when filtering, I look at the properties inside the rule schema. Leaving it inside the JsonRuleEngine and each engine having their own filtering makes the most sense in my opinion (like Johnathon mentioned above).

VeraBE:
Remember to check the Linux build, do you need help with that?

Some tests seem to be failing there, like this one:

```Failed TestRule [262 ms]
  Failed SeverityOneOnly [232 ms]
  Error Message:
   Assert.IsTrue failed. Expected failing rule Ids do not match actual failed lines.  Expected: [TA-000003]  Actual: [TA-000003,TA-000005,TA-000015,TA-000017]
  Stack Trace:
     at Analyzer.Types.ConfigurationTests.TestRunner.TestRule(TestConfiguration ruleExpectations) in /home/runner/work/template-analyzer/template-analyzer/src/Analyzer.Types.ConfigurationTests/TestRunner.cs:line 54```

File: src/Analyzer.Reports/ConsoleReportWriter.cs [hunk 43]
```diff
@@ -24,6 +35,10 @@ public void WriteResults(IEnumerable<Types.IEvaluation> evaluations, IFileInfo t
             {
                 fileMetadata += Environment.NewLine + $"Parameters File: {parametersFile}";
             }
+            if (configurationsFile != null)
+            {
+                fileMetadata += Environment.NewLine + $"Configurations File: {configurationsFile}";
```
VeraBE, line 40: There's also a Console.WriteLine with this information in some other class, wouldn't this end up duplicated in the console?

src/Analyzer.Types.ConfigurationTests/Tests/ExcludeSeverityThree.json
```diff
@@ -0,0 +1,60 @@
+[
+  {
+    "Template": "AppServices",
+    "Configuration": "ExcludeSeverityThree",
+    "ReportedFailures": [
+      {
+        "Id": "TA-000001",
+        "Severity": 2
```
VeraBE, line 8: The idea of specifying the severity here is to validate that configurations that change the severity of rules work properly? I don't see a test case that validates that feature

src/Analyzer.Types/ConfigurationDefinition.cs
```diff
@@ -0,0 +1,26 @@
...
+
+using System;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.Templates.Analyzer.Types
+{
+    /// <summary>
+    /// Represents a configuration file written in JSON.
+    /// </summary>
+    public class ConfigurationDefinition
```
VeraBE, line 12: Why isn't severityOverride defined here too?

src/Analyzer.Types.ConfigurationTests/Tests/ExcludeSeverityThree.json
anaismiller: Partially. For the moment it is for either filtering by Id and/or by Severity (which I tested). In the future when re-mapping is supported, that will be used. It also seemed cleaner to keep the values together rather than create a separate object/list, etc.

src/Analyzer.Types/ConfigurationDefinition.cs
anaismiller: Because it is not yet implemented.

src/Analyzer.Types.ConfigurationTests/Tests/ExcludeSeverityThree.json
VeraBE: To me, it sounds a bit redundant and confusing to add the severity there, especially if that feature is not yet supported. Because when the tests are executed you should already have access to the rule information, including the severity, to validate the filtering. Imagine also if we change the severity of a rule, we would need to change every single instance in every test case that mentions it

src/Analyzer.Types/ConfigurationDefinition.cs
VeraBE: Is there a reason why we keep that information in the documentation anyways? I vote for removing it, or at least adding an asterisk also in line 13 to say that it is not supported

docs/customizing-evaluation-outputs.md
```diff
@@ -0,0 +1,45 @@
...
+"inclusions": {
+    "severity": [int], // List of severities to include in results.
+    "ids": ["Id"] // List of ids to include in results. 
+}
+```
+Only rules with matching severities and ids will be included in results. If the inclusions object contains values, then the exclusions object will be ignored. 
+
+### SeverityOverrides Object [[*]](#note)
+**NOTE: `severityOverrides` is not yet supported. 
+The `severityOverrides` is composed of a key-value pairs of id and a new severity value. Below is an example of changing the severities on two rules:
```
JohnathonMohr, line 39: (minor) grammar can be fixed in first sentence.

src/Analyzer.JsonRuleEngine/Schemas/RuleDefinition.cs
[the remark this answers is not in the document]
JohnathonMohr: @anaismiller I can work with you offline to get this sorted out.

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -13,6 +13,7 @@
 using Microsoft.Azure.Templates.Analyzer.Types;
 using Microsoft.Azure.Templates.Analyzer.Reports;
 using Newtonsoft.Json.Linq;
+using System.Reflection;
```
JohnathonMohr, line 16: Unused using.

src/Analyzer.Core/TemplateAnalyzer.cs
```diff
@@ -108,5 +108,50 @@ private static string LoadRules()
...
+        /// <returns>Configuration file path contents if a file exists.</returns>
+        private string GetConfigurationFileContents(FileInfo configurationsFilePath)
+        {
+            try
+            {
+                string configurationFileContents = configurationsFilePath == null ? null : File.ReadAllText(configurationsFilePath.FullName);
+                if (configurationFileContents == null)
+                {
+                    var defaultPath = Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location),
+                        "/Configurations/Configuration.json");
```
JohnathonMohr, line 125: The usefulness of `Path.Combine()` here is so no path separation characters need to be included, and then it's safe across Linux and Windows.  Instead of writing out the path, just include each directory/file name as another argument to the function.
``` suggestion
                    var defaultPath = Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location),
                        "Configurations", "Configuration.json");
```

src/Analyzer.Core.BuiltInRuleTests/README.md
```diff
@@ -21,7 +22,7 @@ For example, a JSON rule like the following:
     }
 }
 ```
-... would have tests defined in a configuration file named **"Tests/SuperSecurityCheck.json"**.
+... would have tests defined in a configuration file named *"Tests/SuperSecurityCheck.json"*.
```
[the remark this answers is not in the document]
JohnathonMohr: Changes to this file seem unrelated to the rest of the PR.

src/Analyzer.Core/TemplateAnalyzer.cs
```diff
@@ -108,5 +108,50 @@ private static string LoadRules()
...
+                        return null;
+                    }
+                }
+
+                Console.WriteLine(Environment.NewLine + Environment.NewLine + $"Configurations File: {configurationsFilePath}");
+                return configurationFileContents;
+            }
+            catch (Exception e)
+            {
+                throw new Exception($"Failed to read configurations file.", e);
```
JohnathonMohr, line 142: throw `TemplateAnalyzerException` instead.

src/Analyzer.Core/TemplateAnalyzer.cs
```diff
@@ -108,5 +108,50 @@ private static string LoadRules()
...
+                        return File.ReadAllText(defaultPath);
+                    }
+                    else
+                    {
+                        return null;
+                    }
+                }
+
+                Console.WriteLine(Environment.NewLine + Environment.NewLine + $"Configurations File: {configurationsFilePath}");
+                return configurationFileContents;
```
JohnathonMohr, line 138: If you change what the `if` statements are looking for above, you can group this code with the relevant logic reading the file.
For example:
``` cs
if (configurationsFilePath != null)
{
    string configurationFileContents = File.ReadAllText(configurationsFilePath.FullName);
    // print file name and return contents
}

// continue in default scenario
var defaultPath = ...
if (exists)
{
    // print name and return contents
}

return null;
```

src/Analyzer.Reports/ConsoleReportWriter.cs
JohnathonMohr: I don't think this should print the configuration file in the file metadata - it would be reprinted for every template analyzed.

src/Analyzer.Reports/ConsoleReportWriter.cs
```diff
@@ -39,6 +54,7 @@ private void OutputToConsole(IEnumerable<Types.IEvaluation> evaluations)
                 {
                     string resultString = GenerateResultString(evaluation);
                     var output = $"{IndentedNewLine}{(evaluation.RuleId != "" ? $"{evaluation.RuleId}: " : "")}{evaluation.RuleDescription}" +
+                        $"{TwiceIndentedNewLine}Severity: {evaluation.Severity}" +
```
JohnathonMohr, line 57: Match indentation here - either move it out to the same level as others, or indent others as well.

src/Analyzer.Reports/ConsoleReportWriter.cs
```diff
@@ -15,6 +15,17 @@ public class ConsoleReportWriter : IReportWriter
         internal static string IndentedNewLine = Environment.NewLine + "\t";
         internal static string TwiceIndentedNewLine = Environment.NewLine + "\t\t";
 
+        private IFileInfo configurationsFile;
+
+        /// <summary>
+        /// Constructor of the SarifReportWriter class
+        /// </summary>
+        /// <param name="configurationsFilePath">The configuration file that will be used</param>
+        public ConsoleReportWriter(IFileInfo configurationsFilePath = null)
```
JohnathonMohr, line 24: There might not be a need to store the configuration file path.

src/Analyzer.Reports/SarifReportWriter.cs
```diff
@@ -33,8 +33,9 @@ public class SarifReportWriter : IReportWriter
         /// Constructor of the SarifReportWriter class
         /// </summary>
         /// <param name="reportFile">File where the report will be written</param>
+        /// <param name="configurationsFilePath">The configuration file that will be used</param>
         /// <param name="targetPath">The directory that will be analyzed</param>
-        public SarifReportWriter(IFileInfo reportFile, string targetPath = null)
+        public SarifReportWriter(IFileInfo reportFile, IFileInfo configurationsFilePath = null, string targetPath = null)
```
JohnathonMohr, line 38: Can the config file be removed?

src/Analyzer.Reports/ConsoleReportWriter.cs
anaismiller: I believe you mentioned adding it to the Sarif or something?

src/Analyzer.Core/TemplateAnalyzer.cs
anaismiller: I re-ordered it.

src/Analyzer.Types/ConfigurationDefinition.cs
anaismiller: added. Leaving in doc as it was part of the design

src/Analyzer.JsonRuleEngine/Schemas/RuleDefinition.cs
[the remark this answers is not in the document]
anaismiller: Changed, thank you

anaismiller:
converting to draft - will improve the current tests flow

README.md
```diff
@@ -33,9 +33,9 @@ Argument | Description
 `<template-path>` | The ARM template to analyze
 `<directory-path>` | The directory to find ARM templates (recursively finds all templates in the directory and its subdirectories.)
 **(Optional)** `-p` or `--parameters-file-path` | A [parameters file](https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/parameter-files)
+**(Optional)** `-c` or `--config-file-path` | A custom [configuration file](docs/customizing-evaluation-outputs.md) with custom settings. If the configuration file path argument is not provided, by default, the analyzer will check to see if a file is present in  "ExecutablePath/Configurations/Configuration.json".
```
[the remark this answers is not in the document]
anaismiller: A configuration file is not necessary.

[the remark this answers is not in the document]
anaismiller: I updated the suggested change.

src/Analyzer.Core.UnitTests/TemplateAnalyzerTests.cs
```diff
@@ -149,5 +149,14 @@ public void Create_MissingRulesFile_ThrowsException()
...
+
+        [TestMethod]
+        [ExpectedException(typeof(TemplateAnalyzerException))]
+        public void FilterRules_ConfigurationPathIsInvalid_ThrowTemplateAnalyzerException()
+        {
+            var erroredPath = new FileInfo(Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location),
+                        "NoFile.json"));
+            templateAnalyzer.FilterRules(erroredPath);
+        }
     }
```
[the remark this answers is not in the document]
anaismiller: I added them in the JsonRuleEngineTests as the actual Filtering happens in JsonRuleEngine. The tests would look the same. Do you still recommend I add positive tests in this file now?

src/Analyzer.Types.ConfigurationTests/Tests/ExcludeSeverityThree.json
VeraBE: @anaismiller thoughts on this?

VeraBE: @anaismiller @JohnathonMohr @yane3628 thoughts on this?

VeraBE (approved):

src/Analyzer.Types.ConfigurationTests/Analyzer.Types.ConfigurationTests.csproj
```diff
@@ -0,0 +1,20 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
JohnathonMohr, line 1: This looks to be left by mistake.

src/Analyzer.Types.ConfigurationTests/Tests/ExcludeSeverityThree.json
JohnathonMohr: @VeraBE, @anaismiller and I discussed offline.  I don't think these tests are here anymore, so I think your concern is addressed.

src/Analyzer.Types.ConfigurationTests/Analyzer.Types.ConfigurationTests.csproj
anaismiller: Great catch

src/Analyzer.Core/TemplateAnalyzer.cs
```diff
@@ -101,12 +101,60 @@ public IEnumerable<IEvaluation> AnalyzeTemplate(string template, string paramete
...
+            try
+            {
+                string path;
+                if (configurationsFilePath != null)
+                {
+                    path = configurationsFilePath.FullName;
+                }
+                else
+                {
+                    path = Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location),
```
yane3628, line 132: Missing tests for when 
1. `configurationFilePath` is null
2. Default configuration file is not present

anaismiller: 1. I tested it in JsonRuleEngineTests.FilterRules_ValidInputValues_ReturnCorrectFilteredRules, but makes sense to have it here too as it is one step prior. I added in a test for it. The way that makes more sense to me to test this is if the # of rules does not change, but that would involve changing a lot to make those values accessible to this class.
2. It is implied by (1). 
The test I do not have is if it is in the default location, but then the no configuration FilterRules test would not work (as there is a configuration) and the path with it being a test would get messy.

src/Analyzer.Core.UnitTests/TemplateAnalyzerTests.cs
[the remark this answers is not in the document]
yane3628: I think if you can mock the jsonRuleEngine.FilterRules function then yes. At that point you'd be testing TemplateAnalyzer.FilterRules and not the jsonRuleEngine one. Does that make sense?

src/Analyzer.PowerShellRuleEngine/PowerShellRuleEvaluation.cs
```diff
@@ -23,6 +23,9 @@ public class PowerShellRuleEvaluation : IEvaluation
         /// <inheritdoc/>
         public string HelpUri { get; }
 
+        /// <inheritdoc/>
+        public Severity Severity { get; }
```
JohnathonMohr, line 27: Since this is never set (because TTK doesn't have any severity assignments), shall we assign this an explicit default value?  Say Medium?
``` suggestion
        public Severity Severity { get; } = Severity.Medium
```

src/Analyzer.JsonRuleEngine/Schemas/RuleDefinition.cs
```diff
@@ -35,6 +36,18 @@ internal class RuleDefinition
         [JsonProperty]
         public string HelpUri { get; set; }
 
+        ///// <summary>
+        ///// Gets or sets the severity of the rule.
+        ///// </summary>
+        //[JsonProperty]
+        //public int Severity { get; set; } = 2;
```
JohnathonMohr, line 43: Remove

src/Analyzer.JsonRuleEngine/JsonRuleEvaluation.cs
```diff
@@ -8,6 +8,7 @@
 using Microsoft.Azure.Templates.Analyzer.RuleEngines.JsonEngine.Expressions;
 using Microsoft.Azure.Templates.Analyzer.RuleEngines.JsonEngine.Schemas;
 using Microsoft.Azure.Templates.Analyzer.Types;
+using static Microsoft.Azure.Templates.Analyzer.Types.IEvaluation;
```
JohnathonMohr, line 11: I think this isn't needed.

src/Analyzer.JsonRuleEngine.UnitTests/JsonRuleEngineTests.cs
```diff
@@ -338,5 +340,89 @@ private string CreateRulesFromEvaluationDefinitions(string[] ruleEvaluationDefin
...
+            }
+            if (excludeSeverities != null || excludeIds != null)
+            {
+                foreach (var r in jsonRuleEngine.RuleDefinitions)
+                {
+                    Assert.IsTrue((excludeSeverities != null && !excludeSeverities.Contains(r.Severity)) ||
+                        (excludeIds!= null && !excludeIds.Contains(r.Id)));
+                }
+            }
+        }
```
JohnathonMohr, line 426: Please also add tests for *severityOverrides*

JohnathonMohr (changes requested):

src/Analyzer.Core/TemplateAnalyzer.cs
```diff
@@ -101,12 +101,60 @@ public IEnumerable<IEvaluation> AnalyzeTemplate(string template, string paramete
...
+                        return null;
+                    }
+                }
+
+                Console.WriteLine(Environment.NewLine + Environment.NewLine + $"Configurations File: {path}");
+                return File.ReadAllText(path);
+            }
+            catch (Exception e)
+            {
+                throw new TemplateAnalyzerException($"Failed to read configurations file, but path exists.", e);
```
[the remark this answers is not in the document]
anaismiller: updated again

src/Analyzer.Core/TemplateAnalyzer.cs
```diff
@@ -101,12 +101,60 @@ public IEnumerable<IEvaluation> AnalyzeTemplate(string template, string paramete
...
+                        return null;
+                    }
+                }
+
+                Console.WriteLine(Environment.NewLine + Environment.NewLine + $"Configurations File: {path}");
+                return File.ReadAllText(path);
+            }
+            catch (Exception e)
+            {
+                throw new TemplateAnalyzerException($"Failed to read configurations file.", e);
```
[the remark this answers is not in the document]
anaismiller: updated again

src/Analyzer.JsonRuleEngine.UnitTests/JsonRuleEngineTests.cs
```diff
@@ -338,5 +340,89 @@ private string CreateRulesFromEvaluationDefinitions(string[] ruleEvaluationDefin
...
+            }", 10, new int[] { 3 }, new string[] { "TA-000001" }, DisplayName = "Only Severity 3 and  Id TA-000001")]
+        [DataRow(@"{
+                ""inclusions"": {
+                        ""ids"": [""TA-000001""]
+                    }, 
+                ""severityOverrides"": {
+                    ""TA-000001"": 3
+                    }
+                }", 1, null, new string[] { "TA-000001" }, null, null, DisplayName = "Only Id TA-000001, change Severity to 3")]
+        public void FilterRules_ValidInputValues_ReturnCorrectFilteredRules(string configuration, int expectedEvaluationCount,
```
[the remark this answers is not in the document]
anaismiller: Good call! Added

src/Analyzer.JsonRuleEngine.UnitTests/JsonRuleEngineTests.cs
```diff
@@ -338,5 +340,89 @@ private string CreateRulesFromEvaluationDefinitions(string[] ruleEvaluationDefin
...
+                ""severityOverrides"": {
+                    ""TA-000001"": 3
+                    }
+                }", 1, null, new string[] { "TA-000001" }, null, null, DisplayName = "Only Id TA-000001, change Severity to 3")]
+        public void FilterRules_ValidInputValues_ReturnCorrectFilteredRules(string configuration, int expectedEvaluationCount, 
+            IEnumerable<Severity> includeSeverities = null, IEnumerable<string> includeIds = null, 
+            IEnumerable<Severity> excludeSeverities = null, IEnumerable<string> excludeIds = null)
+        {
+            // Arrange
+            var jsonRuleEngine = JsonRuleEngine.Create(TemplateAnalyzer.LoadRules(), t => null);
```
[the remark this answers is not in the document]
anaismiller: I had a feeling you were going to say this, but wanted to give it a try anyway

src/Analyzer.JsonRuleEngine.UnitTests/JsonRuleEngineTests.cs
anaismiller: I did already - test called "Only Id TA-000001, change Severity to 3"

JohnathonMohr: Ok - but the test never actually validates the severity of the rule anywhere.  In other words, *severityOverrides* could do nothing, and the test would still pass.

src/Analyzer.Core.UnitTests/TemplateAnalyzerTests.cs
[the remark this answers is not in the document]
anaismiller: I have created issue #233 to track this request.

JohnathonMohr (approved):

docs/authoring-json-rules.md
```diff
@@ -21,7 +21,7 @@ Here are the fields that make up a rule definition.
 - The `id` should look like `TA-NNNNNN`, with `NNNNNN` being the next unused number according to the [rule ids already defined](https://github.com/Azure/template-analyzer/blob/main/docs/built-in-bpa-rules.md).
 - The `recommendation` should provide clear but concise guidance on how to modify a template if the rule fails.  If some details are somewhat complex, or the rule takes a bit more to understand, add those details to a guide accessible at the URI in `helpUri`.
 - The `helpUri` is optional, but it is good practice to include.  For built-in rules, this will point to a guide in the GitHub repository.
-- The `severity` is optional. If no severity is provided, it evalutes to 2 by default.
+- The `severity` is optional. If no severity is provided, it defaults to a severity of 2.
```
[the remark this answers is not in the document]
nonik0: Yes

yane3628 (approved):

Issue #199, opened 2021-11-15
Message string for each result in SARIF report.


yongyan-gh:
Please consider creating message strings for each instance of a rule violation (result). 
A message string should contain enough information to enable a user to resolve the problem. It can include dynamic information indicates the instance of a rule violation. e.g. "{0} in the '{1}' organization’s '{2}' project are currently exposed to builds for repository forks."
A tutorial about message string can be found at https://github.com/microsoft/sarif-tutorials/blob/main/docs/Authoring-rule-metadata-and-result-messages.md#message-strings

JohnathonMohr:
PowerShell results especially don't play well with SARIF right now because we're limited by the strings the ARM TTK outputs for descriptions, and the SARIF writer assumes each rule will have a standard description that's constant for all violations, in addition to a possible instance-specific message for failures.

- The JSON rules have a standard description that's the same for all rules, but lacks instance-specific messaging.
- The TTK has instance-specific messaging, but no standard description (failures simply write to the console output, and BPA captures those messages, which contain details of the specific failures).
  - This means the first violation of a given PowerShell rule becomes the "standard description" for all other violations, which is really confusing when the description contains mentions of specific items in the template that only apply to the first violation.

Some thought is needed in designing a messaging system that will work for rules in general (JSON rules can use it immediately). The PowerShell rule messaging can be updated to work properly once a proper PowerShell engine is implemented (that isn't just a wrapper around TTK).

PR #204, opened 2021-11-19
Fix SARIF reporting with PowerShell results

### LINKED ISSUE #203: SARIF result writer throws exceptions writing results from PowerShell Engine
If there are PowerShell rule violations and outputting to SARIF is specified on the CLI (especially in Linux), exceptions are thrown for each PowerShell rule violation, and none get written to the SARIF.

This is happening because the PowerShell `Evaluation`s don't populate Recommendation, and the SARIF writer isn't checking for null before working with the string.

JohnathonMohr:
PowerShell results don't play well with SARIF right now because we're limited by the strings the ARM TTK outputs for descriptions, and the SARIF writer assumes each rule will have a standard description that's constant for all violations, in addition to a possible instance-specific message for failures.
- The JSON rules have a standard description that's the same for all rules, but lacks instance-specific messaging.
- The TTK has instance-specific messaging, but no standard description (failures simply write to the console output, and BPA captures those messages, which contain details of the specific failures).
  - This means the first violation of a given PowerShell rule becomes the "standard description" for all other violations, which is really confusing when the description contains mentions of specific items in the template that only apply to the first violation.

SARIF result writer would also throw an exception when writing results from PowerShell rule execution because `IEvaluation.Recommendation` was null.  This fixes that (closes #203), though PS violations still don't play well in SARIF.

To smooth out these wrinkles while we work on a long-term solution, this PR has the following changes:
- Puts PowerShell rule execution behind a CLI flag, so it's disabled by default
- Add `null`-check in the SARIF result writer
- Populate `PowerShellEvaluation.Recommendation` with an empty string for safety
- Populate `PowerShellEvaluation.RuleId` with something simple for now
- Include rule recommendation text in console output if it's available

To speed up release of this fix, this also includes a version increment to `0.1.1`

### DIFF
--- a/src/Analyzer.Core.UnitTests/TemplateAnalyzerTests.cs
+++ b/src/Analyzer.Core.UnitTests/TemplateAnalyzerTests.cs
@@ -58,6 +58,31 @@ namespace Microsoft.Azure.Templates.Analyzer.Core.UnitTests
             }
         }
 
+        [TestMethod]
+        public void AnalyzeTemplate_NotUsingPowerShell_NoPowerShellViolations()
+        {
+            // Arrange
+            string[] resourceProperties = {
+                GenerateResource(
+                    @"{ ""azureActiveDirectory"": { ""tenantId"": ""tenantIdValue"" } }",
+                    "Microsoft.ServiceFabric/clusters", "resource1")
+            };
+            string template = GenerateTemplate(resourceProperties);
+
+            var evaluations = templateAnalyzer.AnalyzeTemplate(
+                template,
+                templateFilePath: @"..\..\..\..\Analyzer.PowerShellRuleEngine.UnitTests\templates\error_without_line_number.json", // This file has violations from TTK rules
+                usePowerShell: false);
+
+            var evaluationsWithResults = evaluations.ToList().FindAll(evaluation => evaluation.HasResults); // EvaluateRulesAgainstTemplate will always return at least an evaluation for each built-in rule
+
+            // There should be no evaluations with failures because TTK should not have run
+            foreach(IEvaluation evaluation in evaluationsWithResults)
+            {
+                Assert.IsTrue(evaluation.Passed);
+            }
+        }
+
         private string GenerateTemplate(string[] resourceProperties)
         {
             return string.Format(@"{{
--- a/src/Analyzer.Core.UnitTests/TemplateAnalyzerTests.cs
+++ b/src/Analyzer.Core.UnitTests/TemplateAnalyzerTests.cs
@@ -6,6 +6,7 @@ using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Runtime.InteropServices;
+using Microsoft.Azure.Templates.Analyzer.RuleEngines.PowerShellEngine;
 using Microsoft.Azure.Templates.Analyzer.Types;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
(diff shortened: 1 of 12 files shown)


src/Analyzer.Core.UnitTests/TemplateAnalyzerTests.cs
```diff
@@ -58,6 +58,31 @@ public void AnalyzeTemplate_ValidInputValues_ReturnCorrectEvaluations(string res
             }
         }
 
+        [TestMethod]
+        public void AnalyzeTemplate_NotUsingPowerShell_NoPowerShellViolations()
```
VeraBE, line 62: Maybe it's worth adding a test with the same template where there is a failure because TTK checks did run?

src/Analyzer.PowerShellRuleEngine.UnitTests/PowerShellRuleEngineTests.cs
```diff
@@ -131,13 +135,16 @@ public void AnalyzeTemplate_MissingTTKRepository_DoesNotThrowAnException()
 
             var powerShellRuleEngineWrongPath = new PowerShellRuleEngine();
             var evaluations = powerShellRuleEngineWrongPath.AnalyzeTemplate(templateContext);
-            Assert.AreEqual(0, evaluations.Count());
-
-            System.IO.Directory.Move(wrongTTKFolderName, TTKFolderName);
 
-            var powerShellRuleEngine = new PowerShellRuleEngine();
-            evaluations = powerShellRuleEngine.AnalyzeTemplate(templateContext);
-            Assert.AreEqual(1, evaluations.Count());
```
VeraBE, line 140: Can we add back this sanity check somewhere in the test case to validate that we didn't get 0 results because TTK is not reporting anything about this file?

anaismiller (approved):

src/Analyzer.PowerShellRuleEngine/PowerShellRuleEngine.cs
```diff
@@ -85,8 +85,10 @@ public IEnumerable<IEvaluation> AnalyzeTemplate(TemplateContext templateContext)
                     {
                         evaluationResults.Add(new PowerShellRuleResult(false, lineNumber));
                     }
+
+                    var ruleId = (executionResult.Name as string)?.Replace(" ", "") ?? string.Empty;
                     var ruleDescription = executionResult.Name + ". " + uniqueError.Key;
-                    evaluations.Add(new PowerShellRuleEvaluation("", ruleDescription, false, evaluationResults));
+                    evaluations.Add(new PowerShellRuleEvaluation(ruleId, ruleDescription, false, evaluationResults));
```
VeraBE, line 91: Should we add an issue for the "longer-term" solution? To cover fixing the rule ids, rule description, and handling instance-specific information for both engines? I can add it

VeraBE (approved):

src/Analyzer.Core.UnitTests/TemplateAnalyzerTests.cs
JohnathonMohr: Yes, that test is the last `DataRow` of the above test case.

src/Analyzer.PowerShellRuleEngine.UnitTests/PowerShellRuleEngineTests.cs
JohnathonMohr: I removed it because that same file is analyzed in the second `DataRow` of the first test case, so it was redundant here.

src/Analyzer.Core.UnitTests/TemplateAnalyzerTests.cs
VeraBE: I was thinking of the case where we make changes such that that template doesn't trigger errors anymore. We will realize that we need to update the above test case, but we might forget that it's affecting this one too

src/Analyzer.PowerShellRuleEngine/PowerShellRuleEngine.cs
JohnathonMohr: It's covered in #199.

src/Analyzer.PowerShellRuleEngine.UnitTests/PowerShellRuleEngineTests.cs
VeraBE: Same comment as in the other thread

src/Analyzer.Core.UnitTests/TemplateAnalyzerTests.cs
JohnathonMohr: That's fair.  I will add the case in this same test method to ensure it's obvious what needs to change in the event of a test failure.

src/Analyzer.PowerShellRuleEngine.UnitTests/PowerShellRuleEngineTests.cs
JohnathonMohr: I disagree in this case - the intent of the test is not to ensure the PS engine works; it's to ensure the engine doesn't throw an exception if the TTK folder is missing.  Once that's validated, the test can end, and different tests can ensure it actually works as intended.

PR #219, opened 2022-03-03
Fix SARIF report aritifactLocation file path issue

yongyan-gh:
# Description

When scanning a directory containing template files in subfolders, the SARIF report results' location Uri contain only file name, subfolder path is missed.
For example:
-	there is a file with violations at ‘c:\scanDir\intermediateDir\file.json’
-	the cli is invoked to scan directory ‘c:\scanDir’
-	the uri of a result is ‘file.json’,	which should be 'intermediateDir\file.json'

### DIFF
--- a/src/Analyzer.Reports.UnitTests/SarifReportWriterTests.cs
+++ b/src/Analyzer.Reports.UnitTests/SarifReportWriterTests.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.IO.Abstractions;
 using System.Linq;
+using System.Runtime.InteropServices;
 using System.Text;
 using FluentAssertions;
 using Microsoft.CodeAnalysis.Sarif;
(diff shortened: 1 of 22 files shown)


yongyan-gh:
@JohnathonMohr pls review the fix

anaismiller (approved):

src/Analyzer.Reports/SarifReportWriter.cs
```diff
@@ -148,7 +150,7 @@ private void ExtractResult(IEvaluation rootEvaluation, IEvaluation evaluation, s
                                     Uri = new Uri
                                     (
                                         UriHelper.MakeValidUri(filePath),
-                                        UriKind.Relative
+                                        UriKind.RelativeOrAbsolute
                                     ),
                                     UriBaseId = UriBaseIdString,
```
JohnathonMohr, line 155: Shouldn't the `UriBaseId` not be set if `filePath` ends up being a full path, not having a root equal to `uriBaseIdString`, as a result of line 60?  It's possible this could happen when processing nested templates - the nested template could be outside the directory scanned, or possibly even a URL in the future.

JohnathonMohr (approved):
Just one potential issue that could arise later.  If you'd prefer, we can create an issue to instead address it later.

src/Analyzer.Reports/SarifReportWriter.cs
yongyan-gh: Updated code to not set UriBaseId if templates path does not belong to the root path.

src/Analyzer.Reports/SarifReportWriter.cs
```diff
@@ -183,6 +187,16 @@ private void PersistReport(IList<Result> sarifResults)
...
+        internal static bool IsSubPath(string rootPath, string childFilPath)
+        {
+            if (!rootPath.EndsWith(Path.DirectorySeparatorChar))
+            {
+                rootPath += Path.DirectorySeparatorChar;
+            }
+
+            return childFilPath.StartsWith(rootPath, StringComparison.OrdinalIgnoreCase);
+        }
+
```
JohnathonMohr, line 199:
``` suggestion
        internal static bool IsSubPath(string rootPath, string childFilePath)
        {
            if (!rootPath.EndsWith(Path.DirectorySeparatorChar))
            {
                rootPath += Path.DirectorySeparatorChar;
            }

            return childFilePath.StartsWith(rootPath, StringComparison.OrdinalIgnoreCase);
        }
```

JohnathonMohr (changes requested):
Looks like there are some test failures.

yongyan-gh:
> Looks like there are some test failures.

The test case uses windows path failed when it runs in linux platform. Looking into it...

VeraBE (approved):

src/Analyzer.Reports.UnitTests/SarifReportWriterTests.cs
```diff
@@ -40,17 +41,16 @@ public void WriteResults_Evalutions_ReturnExpectedSarifLog()
...
-                new { FilePath = @"\\hostname\share\config.json", RootPath = Path.Combine("c:", "foo"), Expected = false },
+                new { FilePath = Path.Combine(rootDir, "foo", "bar", "config.json"), RootPath = Path.Combine(rootDir, "foo"), Expected = true },
+                new { FilePath = Path.Combine(rootDir, "foo", "bar", "level2", "config.json"), RootPath = Path.Combine(rootDir, "foo"), Expected = true },
+                new { FilePath = Path.Combine(rootDir, "foo", "config.json"), RootPath = Path.Combine(rootDir, "foo", "bar"), Expected = false },
+                new { FilePath = Path.Combine(rootDir, "foo", "bar", "config.json"), RootPath = Path.Combine(rootDir, "foo", ""), Expected = true },
+                new { FilePath = Path.Combine(rootDir, "anotherPath", "foo", "bar", "config.json"), RootPath = Path.Combine(rootDir, "foo"), Expected = false },
+                new { FilePath = "https://example.com/config.json", RootPath = Path.Combine(rootDir, "foo"), Expected = false },
+                new { FilePath = @"\\hostname\share\config.json", RootPath = Path.Combine(rootDir, "foo"), Expected = false },
             };
```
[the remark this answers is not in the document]
yongyan-gh: yes, i was looking into that, for some OS user can turn on/off file system case sensitivity, so only like at the OS may not be enough to tell if the current OS file system is case sensitive or not.

A helper function should be able to tell file system case sensitivity:
```
public static bool IsFileSystemCaseSensitive() {
   var tmp = Path.GetTempPath();
   return !Directory.Exists(tmp.ToUpper()) || !Directory.Exists(tmp.ToLower());
}
```
What do you think?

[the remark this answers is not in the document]
JohnathonMohr: Good catch.  Sounds good to me.

JohnathonMohr (approved):

PR #238, opened 2022-04-15
Fixed bug: Exception thrown on JSON files with top-level Arrays #234

### LINKED ISSUE #234: [BUG] Exception thrown on JSON files with top-level Arrays
### Describe the bug

When the CLI checks a JSON file to see if it's a valid template (`IsValidSchema()`), it attempts to parse the file as a `JObject` to see if a *$schema* property is defined.  However, this doesn't work if the top-level JSON structure in the file is a `JArray` instead, resulting in an exception and a cryptic message printed on the CLI:
> An exception occurred: Error reading JObject from JsonReader. Current JsonReader item is not an object: StartArray. Path '', line 1, position 1.

### Expected behavior

The CLI should be able to handle any content in a JSON file (even improperly formatted) and simply skip the file if it's not a template (not having a *$schema* property, or the value of that property not representing an ARM template).  In the case of a top-level `JArray`, this is not in the correct format for an ARM template, so these files can be skipped.

### Reproduction Steps

Create a simple JSON file with an array as the starting structure.  For example:
```
[ ]
```

Attempt to scan that file with the CLI.  It will throw an exception and print the message described above.

### Environment

_No response_

rvbnr-a:
## Description
Fixes #234. This commit updates the IsValidSchema function and renames it to IsValidTemplate. This function reads the file dynamically looking for the schema property. If the property is not found, the input properties will be checked against an array of expected ARM template properties. If neither the schema nor expected properties are found, the file will not be considered a valid arm template.

---

- [ ] Pull request includes test coverage for the included changes.


JohnathonMohr:
@rvbnr-a can you:
- Add a description the to pull request summary (under the 'Description' header)
- Add [closing keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) to automatically link the associated issue to the PR

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -9,10 +13,6 @@
 using System.IO.Abstractions;
 using System.Linq;
 using System.Threading.Tasks;
-using Microsoft.Azure.Templates.Analyzer.Core;
-using Microsoft.Azure.Templates.Analyzer.Types;
-using Microsoft.Azure.Templates.Analyzer.Reports;
-using Newtonsoft.Json.Linq;
```
JohnathonMohr, line 15: Please revert the change in ordering.

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -114,6 +114,7 @@ private RootCommand SetupCommandLineAPI()
 
         private int AnalyzeTemplate(FileInfo templateFilePath, FileInfo parametersFilePath, ReportFormat reportFormat, FileInfo outputFilePath, bool runTtk, bool printMessageIfNotTemplate = true, IReportWriter writer = null)
         {
+
```
JohnathonMohr, line 117: Remove extra newline

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -245,15 +246,32 @@ private void FindJsonFilesInDirectoryRecursive(DirectoryInfo directoryPath, List
 
         private bool IsValidSchema(string template)
         {
-            JObject jsonTemplate = JObject.Parse(template);
-            string schema = (string)jsonTemplate["$schema"];
+            JsonTextReader reader = new JsonTextReader(new StringReader(template));
+            //Check for start object token
```
JohnathonMohr, line 250: Add empty line to give space between code blocks.

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -245,15 +246,32 @@ private void FindJsonFilesInDirectoryRecursive(DirectoryInfo directoryPath, List
...
-            JObject jsonTemplate = JObject.Parse(template);
-            string schema = (string)jsonTemplate["$schema"];
+            JsonTextReader reader = new JsonTextReader(new StringReader(template));
+            //Check for start object token 
+            reader.Read();
+            if (reader.TokenType != JsonToken.StartObject)
+            {
+                return false;   
+            }
+            // Checks that token type is property name, property name is string and property name value is "$schema"
```
JohnathonMohr, line 256: Add empty line above.

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -285,5 +303,7 @@ private IReportWriter GetReportWriter(string reportFormat, FileInfo outputFile,
             }
             return new ConsoleReportWriter();
         }
+
+
```
JohnathonMohr, line 307: Remove extra lines

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -245,15 +246,32 @@ private void FindJsonFilesInDirectoryRecursive(DirectoryInfo directoryPath, List
...
+            {
+                return false;   
+            }
+            // Checks that token type is property name, property name is string and property name value is "$schema"
+            reader.Read();
+            if (reader.TokenType != JsonToken.PropertyName || (string)reader.Value != "$schema")
+            {
+                return false;
+            }
+            //Checks that passed in schema value is among valid schema array
```
JohnathonMohr, line 262: Add empty line above.

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -245,15 +246,32 @@ private void FindJsonFilesInDirectoryRecursive(DirectoryInfo directoryPath, List
...
+                return false;
+            }
             string[] validSchemas = { 
                 "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
                 "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
                 "https://schema.management.azure.com/schemas/2018-05-01/subscriptionDeploymentTemplate.json#",
                 "https://schema.management.azure.com/schemas/2019-08-01/tenantDeploymentTemplate.json#",
                 "https://schema.management.azure.com/schemas/2019-08-01/managementGroupDeploymentTemplate.json#"};
-            return validSchemas.Contains(schema);
+           return validSchemas.Contains(reader.Value);
```
JohnathonMohr, line 274: Fix indenting (one space was removed in front)

JohnathonMohr (changes requested):
A few code styling adjustments

JohnathonMohr:
One issue with how this change checks for proper schema version is that it will only work for templates which have `$schema` as the first property in the template.  This is the case for probably almost all templates, but I'm a little concerned it will skip templates that simply reorder that property (which is also an "easy" way for someone to author a template and prevent it from being scanned).

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -245,15 +246,32 @@ private void FindJsonFilesInDirectoryRecursive(DirectoryInfo directoryPath, List
 
         private bool IsValidSchema(string template)
         {
-            JObject jsonTemplate = JObject.Parse(template);
-            string schema = (string)jsonTemplate["$schema"];
+            JsonTextReader reader = new JsonTextReader(new StringReader(template));
```
VeraBE, line 249:
```suggestion
            var reader = new JsonTextReader(new StringReader(template));
```

JohnathonMohr:
> One issue with how this change checks for proper schema version is that it will only work for templates which have `$schema` as the first property in the template. This is the case for probably almost all templates, but I'm a little concerned it will skip templates that simply reorder that property (which is also an "easy" way for someone to author a template and prevent it from being scanned).

I have one idea for how this can be done, balancing performance (only read what's necessary) and completeness (not skipping templates).  I will share if interested.

src/Analyzer.Cli/CommandLineParser.cs
yane3628: > System.* namespaces should be listed first

See https://github.com/Azure/template-analyzer/blob/development/CONTRIBUTING.md#coding-conventions

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -243,23 +243,66 @@ private void FindJsonFilesInDirectoryRecursive(DirectoryInfo directoryPath, List
...
+                return false;
+            }
+
+            // Checks all top level properties for schema property. If schema is not found, will check property names against expected ARM template properties. If expected properties
+            // are not found function will return false.
+            while (reader.Read())
+            {
+                if (reader.Depth == 1 && reader.TokenType == JsonToken.PropertyName)
+                {
+                    string[] expectedProperties =
```
JohnathonMohr, line 262: Rename to something a little more descriptive, like `validTemplateProperties`.

JohnathonMohr, line 262: Instead of initializing this array for every iteration, create it once outside the loop.  This can be either just before the `while` loop, or as part of the class just above the constructor.

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -243,23 +243,66 @@ private void FindJsonFilesInDirectoryRecursive(DirectoryInfo directoryPath, List
...
+                    if (reader.Value.ToString() == "$schema")
+                    {
+                        reader.Read();
+
+                        if (reader.TokenType != JsonToken.String)
+                        {
+                            return false;
+                        }
+
+                        string[] validSchemas = {
```
JohnathonMohr, line 282: Now that this is inside a loop, create this outside the loop too, like the properties array.

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -243,23 +243,66 @@ private void FindJsonFilesInDirectoryRecursive(DirectoryInfo directoryPath, List
...
+                            "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+                            "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
+                            "https://schema.management.azure.com/schemas/2018-05-01/subscriptionDeploymentTemplate.json#",
+                            "https://schema.management.azure.com/schemas/2019-08-01/tenantDeploymentTemplate.json#",
+                            "https://schema.management.azure.com/schemas/2019-08-01/managementGroupDeploymentTemplate.json#"
+                        };
+
+                        return validSchemas.Contains(reader.Value);
+                    }
+                    else if (!expectedProperties.Contains(reader.Value))
```
JohnathonMohr, line 292: ARM templates are not case-sensitive.  So if a template has a resources array with the key `Resources` instead of `resources`, this check will fail.  I suggest you do a case-insensitive comparison.
```suggestion
                    else if (!expectedProperties.Any(property => property.Equals(reader.Value, StringComparison.OrdinalIgnoreCase)))
```

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -243,23 +243,66 @@ private void FindJsonFilesInDirectoryRecursive(DirectoryInfo directoryPath, List
...
+
+                        string[] validSchemas = {
+                            "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+                            "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
+                            "https://schema.management.azure.com/schemas/2018-05-01/subscriptionDeploymentTemplate.json#",
+                            "https://schema.management.azure.com/schemas/2019-08-01/tenantDeploymentTemplate.json#",
+                            "https://schema.management.azure.com/schemas/2019-08-01/managementGroupDeploymentTemplate.json#"
+                        };
+
+                        return validSchemas.Contains(reader.Value);
```
JohnathonMohr, line 290: This wasn't doing a case-insensitive comparison before, but it would be good to update it now.  You can do it similarly to the suggestion for property comparison.

JohnathonMohr (changes requested):

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -243,23 +243,69 @@ private void FindJsonFilesInDirectoryRecursive(DirectoryInfo directoryPath, List
...
+                    if (reader.Value.ToString() == "$schema")
+                    {
+                        reader.Read();
+
+                        if (reader.TokenType != JsonToken.String)
+                        {
+                            return false;
+                        }
+
+                        return validSchemas.Any(property => JsonToken.Equals(reader.Value,StringComparison.OrdinalIgnoreCase));
```
JohnathonMohr, line 293: I don't think this is done right - it probably doesn't compile.  Please double-check this after resolving the conflict.

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -243,23 +243,69 @@ private void FindJsonFilesInDirectoryRecursive(DirectoryInfo directoryPath, List
...
+               "https://schema.management.azure.com/schemas/2019-08-01/tenantDeploymentTemplate.json#",
+               "https://schema.management.azure.com/schemas/2019-08-01/managementGroupDeploymentTemplate.json#"
+            };
+
+            while (reader.Read())
+            {
+                if (reader.Depth == 1 && reader.TokenType == JsonToken.PropertyName)
+                {
+
+
```
JohnathonMohr, line 283: Remove empty lines

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -243,23 +243,68 @@ private void FindJsonFilesInDirectoryRecursive(DirectoryInfo directoryPath, List
...
+               "https://schema.management.azure.com/schemas/2019-08-01/tenantDeploymentTemplate.json#",
+               "https://schema.management.azure.com/schemas/2019-08-01/managementGroupDeploymentTemplate.json#"
+            };
+
+            while (reader.Read())
+            {
+                if (reader.Depth == 1 && reader.TokenType == JsonToken.PropertyName)
+                {
+
+                    if (reader.Value.ToString() == "$schema")
```
JohnathonMohr, line 283: This should be compared with `OrdinalIgnoreCase` too.

src/Analyzer.Cli.FunctionalTests/CommandLineParserTests.cs
```diff
@@ -167,5 +168,32 @@ private static string GetFilePath(string testFileName)
         {
             return Path.Combine(Directory.GetCurrentDirectory(), "Tests", testFileName);
         }
+
+        [DataTestMethod]
+        [DataRow(ValidTemplateConstants.PassingTest, 0, DisplayName = "Valid Template")]
+        [DataRow(ValidTemplateConstants.CaseSensitivity, 1, DisplayName = "Unexpected capitalization in schema")]
+        [DataRow(ValidTemplateConstants.DifferentSchemaDepths, 1, DisplayName = "Two schemas, different depths, valid schema last")]
```
JohnathonMohr, line 175: These tests are misleading.  In both of these cases, `IsValidTemplate` should (and likely does) return `true`.  The exit code returned by analyze-template would be a result of the test text not being a valid template, and is not related to the schema, which is what this test should be validating.

For these cases, I suggest making the text a valid template that results in a 0 exit code, so it demonstrates that `IsValidTemplate` gave the correct results.

Additionally, the description "Unexpected capitalization in schema" makes it sound like this scenario isn't allowed, but the logic does allow for it by design.  Can you rephrase so it's clear that this is a valid scenario?

src/Analyzer.Cli.FunctionalTests/ValidTemplateConstants.cs
```diff
@@ -0,0 +1,50 @@
...
+        public const string MissingSchema = @"{""contentVersion"": ""1.0.0.0"", ""Parameters"": {}}";
+
+        public const string SchemaValueNotString = @"{ ""$schema"": 5 }";
+
+        public const string NoSchemaInvalidProperties = @" {""properties"": }";
+
+        public const string DifferentSchemaDepths = @"
+        {
+            ""parameters"": ""{
+              ""$schema"": ""https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#""
```
JohnathonMohr, line 31: I believe the idea of this test is to demonstrate that the inner schema is ignored, and the outer schema is what is actually honored.  However, because they are the same thing, there is no way to verify in the test which value was used.  To test this properly, the inner schema should be set to an invalid value so that if the wrong one is taken, the test fails.

Additionally, continuing from my other comment, this should be made into a valid template so it's clear that `IsValidTemplate` returned correctly.

JohnathonMohr (changes requested):
Overall looking pretty good.  Just a few adjustments needed in the tests to make them clearer.

JohnathonMohr (approved):

src/Analyzer.Cli.FunctionalTests/CommandLineParserTests.cs
```diff
@@ -8,10 +8,11 @@
 using Microsoft.Azure.Templates.Analyzer.Cli;
 using Newtonsoft.Json.Linq;
 
+
 namespace Analyzer.Cli.FunctionalTests
 {
     [TestClass]
-    public class CommandLineParserTests
+    public class CommandLineParserTests
```
nonik0, line 15: minor nit: added extra space here

src/Analyzer.Cli.FunctionalTests/CommandLineParserTests.cs
```diff
@@ -167,5 +168,32 @@ private static string GetFilePath(string testFileName)
...
         }
+
+        [DataTestMethod]
+        [DataRow(ValidTemplateConstants.PassingTest, 0, DisplayName = "Valid Template")]
+        [DataRow(ValidTemplateConstants.SchemaCaseInsensitive, 0, DisplayName = "Schema is case insensitive")]
+        [DataRow(ValidTemplateConstants.DifferentSchemaDepths, 0, DisplayName = "Two schemas, different depths, valid schema last")]
+        [DataRow(ValidTemplateConstants.MissingStartObject, 4, DisplayName = "Missing start object")]
+        [DataRow(ValidTemplateConstants.NoValidTopLevelProperties, 4, DisplayName = "Unexpected property depths")]
+        [DataRow(ValidTemplateConstants.MissingSchema, 4, DisplayName = "Missing schema, capitalized property names")]
+        [DataRow(ValidTemplateConstants.SchemaValueNotString, 4, DisplayName = "Schema value isnt string")]
```
nonik0, line 179: Minor nit: "isnt" missing apostrophe

src/Analyzer.Cli.FunctionalTests/ValidTemplateConstants.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Analyzer.Cli.FunctionalTests
+{
+    public class ValidTemplateConstants
+    {
+        public const string MissingStartObject = @" ""$schema"": ""https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#""}";
+
+        public const string NoValidTopLevelProperties =@"
```
nonik0, line 13: Minor nit: missing whitespace

src/Analyzer.Cli.FunctionalTests/ValidTemplateConstants.cs
```diff
@@ -0,0 +1,52 @@
...
+        public const string SchemaCaseInsensitive = @"{""$schEmA"": ""https://ScHeMa.mAnAgEmEnt.AzUrE.cOm/schEmAs/2019-04-01/deploymentTemplate.json#"", ""contentVersion"": ""1.0.0.0"", ""resources"": [] }";
+
+        public const string MissingSchema = @"{""contentVersion"": ""1.0.0.0"", ""Parameters"": {}}";
+
+        public const string SchemaValueNotString = @"{ ""$schema"": 5 }";
+
+        public const string NoSchemaInvalidProperties = @" {""properties"": }";
+ 
+        public const string DifferentSchemaDepths = @"{
+            ""$schema"": ""https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#"",
```
nonik0, line 29: I think the valid schema in this test case should come after the invalid one, i.e. move schema to last property. Otherwise with the way the checks are implemented, the invalid schema won't even be looked at so we don't get the test coverage skipping over it.

src/Analyzer.Cli.FunctionalTests/ValidTemplateConstants.cs
```diff
@@ -0,0 +1,52 @@
...
+            ""contentVersion"": ""1.0.0.0"",
+            ""parameters"": {},
+            ""functions"": [],
+            ""variables"": {""$schema"": ""https://schema.management.azure.com/schemas/2040-40-40/deploymentTemplate.json#""},
```
nonik0, line 33: Minor nit: I think I would prefer a more obvious invalid schema value here, i.e. literally says invalidSchema somewhere or that's the whole value or something like that.

src/Analyzer.Cli.FunctionalTests/ValidTemplateConstants.cs
```diff
@@ -0,0 +1,52 @@
+﻿using System;
```
nonik0, line 1: Minor nit: would be nice to reorder the constants here into the valid and invalid ones, similar to how are separated/ordered in the DataRow attributes in your new test.

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -297,17 +298,62 @@ private void FindJsonFilesInDirectoryRecursive(DirectoryInfo directoryPath, List
...
+            reader.Read();
+            if (reader.TokenType != JsonToken.StartObject)
+            {
+                return false;
+            }
+
+            // Checks all top level properties for schema property. If schema is not found, will check property names against expected ARM template properties. If expected properties
+            // are not found function will return false.
+
+            string[] validTemplateProperties =
```
nonik0, line 314: Minor nit: can this be declared const? Also validSchemas below as well.

nonik0 (changes requested):

src/Analyzer.Cli.FunctionalTests/ValidTemplateConstants.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Analyzer.Cli.FunctionalTests
+{   /// <summary>
+/// Constants being passed in as test cases, first three contants should pass while the following five should fail.
+/// </summary>
```
anaismiller, line 10:
```suggestion
{
    /// <summary>
    /// Constants being passed in as test cases, first three contants should pass while the following five should fail.
    /// </summary>
```

anaismiller: fixing the spacing

src/Analyzer.Cli.FunctionalTests/ValidTemplateConstants.cs
```diff
@@ -0,0 +1,50 @@
...
+    public class ValidTemplateConstants
```
anaismiller, line 11: Are they all Valid? Rename.

src/Analyzer.Cli.FunctionalTests/ValidTemplateConstants.cs
anaismiller: I can imagine in the future the order of the file will change. 
Instead, put the summary as "Constants being passed in as test cases for the CommandLineParserTests"
And on-top of each string, add if it will pass or fail.

anaismiller: also type on "contants"

anaismiller: Or a last option would be to add Pass/Fail at the beginning of the string variable

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -297,17 +298,60 @@ private void FindJsonFilesInDirectoryRecursive(DirectoryInfo directoryPath, List
...
+
+            reader.Read();
+            if (reader.TokenType != JsonToken.StartObject)
+            {
+                return false;
+            }
+
+            // Checks all top level properties for schema property. If schema is not found, will check property names against expected ARM template properties. If expected properties
+            // are not found function will return false.
+            string[] validTemplateProperties = {
```
anaismiller, line 313: Would it be better to make this be a variable at the top of this file?

src/Analyzer.Cli/CommandLineParser.cs
anaismiller: was this resolved?

src/Analyzer.Cli.FunctionalTests/ValidTemplateConstants.cs
```diff
@@ -5,11 +5,13 @@
...
-/// Constants being passed in as test cases, first three contants should pass while the following five should fail.
-/// </summary>
-    public class ValidTemplateConstants
+{   
+    /// <summary>
+    /// Constants being passed in as test cases for the CommandLineParserTests.
+    /// </summary>
+    public class TestcaseTemplateConstants
     {
+        //Pass
```
anaismiller, line 14:
```suggestion
        // Pass
```

src/Analyzer.Cli.FunctionalTests/ValidTemplateConstants.cs
```diff
@@ -20,8 +22,10 @@ public class ValidTemplateConstants
             ""outputs"": {}
         }";
 
+        //Pass
```
anaismiller, line 25:
```suggestion
        // Pass
```

src/Analyzer.Cli.FunctionalTests/ValidTemplateConstants.cs
```diff
@@ -20,8 +22,10 @@ public class ValidTemplateConstants
...
         public const string SchemaCaseInsensitive = @"{""$schEmA"": ""https://ScHeMa.mAnAgEmEnt.AzUrE.cOm/schEmAs/2019-04-01/deploymentTemplate.json#"", ""contentVersion"": ""1.0.0.0"", ""resources"": [] }";
 
+        //Pass
```
anaismiller, line 28:
```suggestion
        // Pass
```

src/Analyzer.Cli.FunctionalTests/ValidTemplateConstants.cs
```diff
@@ -32,19 +36,24 @@ public class ValidTemplateConstants
             ""$schema"": ""https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#""
         }";
 
+        //Fail
```
anaismiller, line 39:
```suggestion
        // Fail
```

src/Analyzer.Cli.FunctionalTests/ValidTemplateConstants.cs
```diff
@@ -32,19 +36,24 @@ public class ValidTemplateConstants
...
         public const string MissingStartObject = @" ""$schema"": ""https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#""}";
-
+        
+        //Fail
```
anaismiller, line 42:
```suggestion
        // Fail
```

src/Analyzer.Cli.FunctionalTests/ValidTemplateConstants.cs
```diff
@@ -32,19 +36,24 @@ public class ValidTemplateConstants
...
         public const string NoValidTopLevelProperties = @"
             { 
               ""parameters"": {
               ""$schema"": ""https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#""
             }    
         }";
 
+        //Fail
```
anaismiller, line 50:
```suggestion
        // Fail
```

src/Analyzer.Cli.FunctionalTests/ValidTemplateConstants.cs
```diff
@@ -32,19 +36,24 @@ public class ValidTemplateConstants
...
         public const string MissingSchema = @"{""contentVersion"": ""1.0.0.0"", ""Parameters"": {}}";
 
+        //Fail
```
anaismiller, line 53:
```suggestion
        // Fail
```

src/Analyzer.Cli.FunctionalTests/ValidTemplateConstants.cs
```diff
@@ -32,19 +36,24 @@ public class ValidTemplateConstants
...
         public const string SchemaValueNotString = @"{ ""$schema"": 5 }";
 
+        //Fail
```
anaismiller, line 56:
```suggestion
        // Fail
```

anaismiller (approved):
Approving, with the suggestions for adding a space

nonik0 (approved):

PR #229, opened 2022-03-17
Add error codes to CLI

anaismiller:
## Description

If an error occurs while the tool runs, the CLI should return unsuccessful error codes.
Closes #211 

---

This checklist is used to make sure that common guidelines for a pull request are followed.
- [x] **I have read the [contribution guidelines](../CONTRIBUTING.md).**
- [x] The pull request does not introduce breaking changes.

### [General Guidelines](../CONTRIBUTING.md)
- [x] Title of the pull request is clear and informative.
- [x] Description of the pull request is clear and informative.
- [x] I have added myself to the 'assignees'.
- [x] I have added 'linked issues' if relevant.

### [Testing Guidelines](../CONTRIBUTING.md#tests)
- [ ] Pull request includes test coverage for the included changes.

Issue #207, opened 2021-12-06
Extract common functionality between AnalyzeDirectory and AnalyzeTemplate in CLI


JohnathonMohr:
Code duplication is starting to build up between these two methods in the CLI.  The common code should be extracted into something like `AnalysisSetup` and `AnalysisCleanup` methods to perform common setup across scenarios (scanning a template/directory) and then call a method to scan the template.

Something like:
```
AnalyzeTemplateCommandHandler()
    AnalysisSetup()
    AnalyzeTemplate()
    AnalysisCleanup()

AnalyzeDirectoryCommandHandler()
    AnalysisSetup()
    foreach template found...
        AnalyzeTemplate()
    AnalysisCleanup()

AnalyzeTemplate()
    <Scan template and report results>

AnalysisSetup()
    <Common setup between scenarios> // i.e. report initialization, filtering rules, etc.

AnalysisCleanup()
    <Common cleanup> // i.e. finalize reporting, etc.
```

PR #239, opened 2022-04-15
part 2 exitCodes and add Cli Functional Tests

### LINKED ISSUE #28: Add tests for the CLI
The CLI project doesn't have any tests.  Tests need to be written for it.

### LINKED ISSUE #211: Define error scenarios and behaviors for CLI
If an error occurs while the tool runs, the CLI should return unsuccessful error codes.
We need to add this to the documentation.

anaismiller:
## Description

Add exit code returning logic
Add Cli Functional Tests
Closes #211 
Closes #28 

---


src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -235,15 +241,30 @@ private void AnalyzeDirectory(DirectoryInfo directoryPath, ReportFormat reportFo
...
             }
         }
 
+        private int CalculateExitCode(int exitCode, int res)
+        {
+            if (exitCode == 6 || res == 6)
+                return 6;
+            else if ((exitCode == 5 && res >= 1 && res <= 4) || (res == 5 && exitCode >= 1 && exitCode <= 4))
+                return 6;
+            else if (res == 5)
```
VeraBE, line 262: Shouldn't this one be `res == 5 || exitCode == 5`? It would also fall into the next `else`, I don't think this case is necessary actually

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -193,14 +193,14 @@ private void AnalyzeDirectory(DirectoryInfo directoryPath, ReportFormat reportFo
...
                 {
                     logger.LogError("Invalid directory: {directoryPath}", directoryPath);
-                    return;
+                    return 2;
                 }
 
                 // Check that output file path provided for sarif report
                 if (reportFormat == ReportFormat.Sarif && outputFilePath == null)
                 {
                     logger.LogError("Output file path is not provided.");
```
VeraBE, line 202:
```suggestion
                    logger.LogError("Output file path was not provided.");
```

Unrelated to the PR, but I just realized we are outputting two different messages for the same error

anaismiller: I don't see it. Where is the 2nd instance of this error?

src/Analyzer.Cli/CommandLineParser.cs
anaismiller: Error: #s 1 -4
Issue: # 5
Error + Issue: # 6

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -235,15 +241,30 @@ private void AnalyzeDirectory(DirectoryInfo directoryPath, ReportFormat reportFo
...
+        private int CalculateExitCode(int exitCode, int res)
+        {
+            if (exitCode == 6 || res == 6)
+                return 6;
+            else if ((exitCode == 5 && res >= 1 && res <= 4) || (res == 5 && exitCode >= 1 && exitCode <= 4))
+                return 6;
+            else if (res == 5)
+                return 5;
+            else
+                return Math.Max(exitCode, res);
```
anaismiller, line 265: I was not sure if I should do a Max or not. What number should be returned if there are multiple errors?

src/Analyzer.Cli/CommandLineParser.cs
anaismiller, line 262: For the moment, there is no return 5 issue in the code in the CLI. Where do you think we should return 5?

src/Analyzer.Cli/CommandLineParser.cs
VeraBE: Lines 144 and 202

src/Analyzer.Cli/CommandLineParser.cs
VeraBE: For me it makes sense to return 1 in that case

src/Analyzer.Cli/CommandLineParser.cs
anaismiller: `Shouldn't this one be res == 5 || exitCode == 5?` It is implicit. 
Scenario: Exit code is already a 5. a) res is a 0. This function is not called. b) res is 1 - 4. It is covered in above `else if`. c) res is a 5. already covered the way it is written. 
Now for the `I don't think this case is necessary actually`. You are correct. I put it in case we change the end logic. If it stays that way, then I'll remove. Thanks for pointing it out.

VeraBE: Yes, but if by the last else either res or exitCode are 5 then that's what's going to be returned, if you do max

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -235,15 +241,30 @@ private void AnalyzeDirectory(DirectoryInfo directoryPath, ReportFormat reportFo
...
                 }
             }
             catch (Exception exp)
             {
                 logger.LogError(GetExceptionMessage(exp));
+                return 1;
             }
         }
 
+        private int CalculateExitCode(int exitCode, int res)
```
JohnathonMohr, line 256: I think the logic here can be greatly simplified, to the point of maybe not needing this function.  For example, since this function is only called from `AnalyzeDirectory()`, `res` will never be anything other than 0, 1, or 4.  Nothing different is needed for 4, and whether 0s and 1s are returned can be detected at the end by looking at `numOfSuccesses` and `filesFailed.Count`.

src/Analyzer.Cli/CommandLineParser.cs
JohnathonMohr: @VeraBE can you explain?  What are the two messages, for what error?

JohnathonMohr (changes requested):
Can you simplify the return code?  I think it can be easily calculated at the end of AnalyzeDirectory.

src/Analyzer.Cli/CommandLineParser.cs
JohnathonMohr: 144 is only logged if `AnalyzeTemplate()` is not called from `AnalyzeDirectory()`, where `AnalyeDirectory()` will log it instead.  This is another area that will be improved with #207.

src/Analyzer.Cli/CommandLineParser.cs
JohnathonMohr: I think we can remove this function altogether.  See my other comment at the top of this function.

src/Analyzer.Cli/CommandLineParser.cs
anaismiller: Oh, I thought 2 and 3 could happen in `AnalyzeTemplate()`...

JohnathonMohr: 2 and 3 can be returned from that function, but when called from `AnalyzeDirectory()`, they never will be.
- 2 means the file doesn't exist - but the file will always exist in this case because it was discovered from `FindJsonFilesInDirectoryRecursive()`
- 3 will never be returned because `writer` is always passed from `AnalyzeDirectory()`

anaismiller: Thank you for clarifying Johnathon!

anaismiller: Wait so what about when we add in 5? Then would the logic be needed?

JohnathonMohr: Where is 5 initially returned/generated? I don't see it.

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -193,14 +193,14 @@ private void AnalyzeDirectory(DirectoryInfo directoryPath, ReportFormat reportFo
                 if (!directoryPath.Exists)
                 {
                     logger.LogError("Invalid directory: {directoryPath}", directoryPath);
-                    return;
+                    return 2;
```
JohnathonMohr, line 196: Define constants for the exit codes at the top of the file and return those instead of raw numbers.

src/Analyzer.Cli/CommandLineParser.cs
anaismiller: Yes, that is one of my questions

src/Analyzer.Cli/CommandLineParser.cs
anaismiller: sounds good

src/Analyzer.Cli/CommandLineParser.cs
JohnathonMohr: Was `5` never implemented in #229?

src/Analyzer.Cli/CommandLineParser.cs
yane3628: I've seen error codes in other tools where it says "Errors and issues" and "Errors but no issues". Essentially a new code for multiple errors.

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -239,6 +248,11 @@ private void AnalyzeDirectory(DirectoryInfo directoryPath, FileInfo configuratio
                         else if (res == 1)
                         {
                             filesFailed.Add(file);
+                            exitCode = CalculateExitCode(exitCode, res);
```
yane3628, line 251: If we keep the CalculateExitCode functoin, I think ```exitCode``` and ```res``` should be renamed to be more descriptive variables.

src/Analyzer.Cli/CommandLineParser.cs
yane3628: Can it also be simplified by making them different enums (e.g. a set of enums of errors, issues, and success) or something? I'm thinking some other type of data structure where you have more to compare than just ints.

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -143,14 +151,14 @@ private int AnalyzeTemplate(FileInfo templateFilePath, FileInfo parametersFilePa
                 if (!templateFilePath.Exists)
                 {
                     logger.LogError("Invalid template file path: {templateFilePath}", templateFilePath);
-                    return 2;
+                    return ErrorInvalidFile;
```
yane3628, line 154: Can we add tests to test exit codes for the CLI? At least for the easy ones like invalid file/missing file?

yane3628 (changes requested):

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -23,6 +23,14 @@ internal class CommandLineParser
 
         private readonly TemplateAnalyzer templateAnalyzer;
 
+        private const int Success = 0;
```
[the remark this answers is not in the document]
JohnathonMohr: The README already has these.

src/Analyzer.Cli/CommandLineParser.cs
JohnathonMohr: One option is to do bitwise enums and `|` values together to form the final exit code.

src/Analyzer.Cli/CommandLineParser.cs
anaismiller: CLI does not have tests yet. I wanted to add a test proj, but was not sure if it was appropriate to do it in this PR or in a separate PR.

src/Analyzer.Cli/CommandLineParser.cs
anaismiller: Revisiting no longer needed bitwise enums and also thought it would be better for the user that exit codes are sequential

src/Analyzer.Cli/CommandLineParser.cs
anaismiller: declining on "Errors but no issues"

src/Analyzer.Cli/CommandLineParser.cs
yane3628: I think it'd be appropriate to add a test project and some number of tests related to your work in this PR. That way we can the CLI test coverage.

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -143,14 +153,14 @@ private int AnalyzeTemplate(FileInfo templateFilePath, FileInfo parametersFilePa
...
                 {
                     logger.LogError("Invalid template file path: {templateFilePath}", templateFilePath);
-                    return 2;
+                    return (int)ExitCode.ErrorInvalidPath;
                 }
 
                 // Check that output file path provided for sarif report
                 if (writer == null && reportFormat == ReportFormat.Sarif && outputFilePath == null)
                 {
                     logger.LogError("Output file path was not provided.");
```
yane3628, line 162: This isn't part of the scope of your PR, but I want to recommend a different wording for this error message.

yane3628, line 162:
```suggestion
                    logger.LogError("When using --report-format sarif flag, --output-file-path flag is requred.");
```

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -181,12 +191,12 @@ private int AnalyzeTemplate(FileInfo templateFilePath, FileInfo parametersFilePa
 
                 writer.WriteResults(evaluations, (FileInfoBase)templateFilePath, (FileInfoBase)parametersFilePath);
 
-                return 0;
+                return evaluations.Where(e => !e.Passed).Count() > 0 ? (int)ExitCode.Issue : (int)ExitCode.Success;
```
yane3628, line 194:
```suggestion
                return evaluations.Any(e => !e.Passed) ? (int)ExitCode.Issue : (int)ExitCode.Success;
```

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -206,14 +216,14 @@ private void AnalyzeDirectory(DirectoryInfo directoryPath, FileInfo configuratio
...
-                    return;
+                    return (int)ExitCode.ErrorInvalidPath;
                 }
 
                 // Check that output file path provided for sarif report
                 if (reportFormat == ReportFormat.Sarif && outputFilePath == null)
                 {
-                    logger.LogError("Output file path is not provided.");
-                    return;
+                    logger.LogError("Output file path was not provided.");
```
yane3628, line 225:
```suggestion
                    logger.LogError("When using --report-format sarif flag, --output-file-path flag is requred.");
```

src/Analyzer.Cli/CommandLineParser.cs
anaismiller: I'm listening

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -23,6 +23,16 @@ internal class CommandLineParser
...
         private readonly TemplateAnalyzer templateAnalyzer;
 
+        private enum ExitCode {
+            Success = 0,
+            ErrorGeneric = 1,
+            ErrorInvalidPath = 2,
+            ErrorMissingPath = 3,
+            ErrorInvalidARMTemplate = 4,
+            Issue = 5,
+            ErrorIssue = 6,
```
VeraBE, line 33: ErrorsAndIssues might be more declarative?

VeraBE (approved):

src/Analyzer.Cli/Program.cs
```diff
@@ -7,11 +7,11 @@ namespace Microsoft.Azure.Templates.Analyzer.Cli
 {
     class Program
     {
-        static async Task Main(string[] args)
+        static async Task<int> Main(string[] args)
         {
             CommandLineParser commandLineParser = new CommandLineParser();
 
-            await commandLineParser.InvokeCommandLineAPIAsync(args).ConfigureAwait(false);
+           return await commandLineParser.InvokeCommandLineAPIAsync(args).ConfigureAwait(false);
```
VeraBE, line 14:
```suggestion
            return await commandLineParser.InvokeCommandLineAPIAsync(args).ConfigureAwait(false);
```

anaismiller: Nice catch!

src/Analyzer.Cli.UnitTests/AppServicesLogs-Failures.json
```diff
@@ -0,0 +1,89 @@
+{
```
anaismiller, line 1: I was not sure if I should create the file in TestInit() method, then clean them up at the end. I think it is cleaner to have them be created there, but then that means for each test, it would re-create them (sometimes unnecessarily). I decided to then create them and have them be copied to output if newer.

src/Analyzer.Cli.UnitTests/CommandLineParserTests.cs
```diff
@@ -0,0 +1,98 @@
...
+            var templatePath = String.Concat(Directory.GetCurrentDirectory(), "\\AppServicesLogs-Failures.json");
+            var parametersFilePath = String.Concat(Directory.GetCurrentDirectory(), relativeParametersFilePath);
+            var args = new string[] { "analyze-template", templatePath, "--parameters-file-path", parametersFilePath };
+            var result = _commandLineParser.InvokeCommandLineAPIAsync(args);
+
+            Assert.AreEqual(expectedExitCode, result.Result);
+        }
+
+        [TestMethod]
+        public void AnalyzeTemplate_ConfigurationFileParamUsed_ReturnExitCode()
```
anaismiller, line 53: I was not sure if I should bring in Analyzer.Reports.ConsoleReportWriter to be able to read the test outputs (ie. would benefit the ConfigurationFile test). These are functional tests so it should test everything end-to-end, but at the same time, those are already tested in different projects and I did not want to add an additional project dependency to this project. Thoughts?

src/Analyzer.Cli.UnitTests/CommandLineParserTests.cs
```diff
@@ -0,0 +1,98 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Microsoft.Azure.Templates.Analyzer.Cli;
+using System.Linq;
```
yane3628, line 6: System using statements should be at the top of the list.

src/Analyzer.Cli.UnitTests/AppServicesLogs-Failures.json
yane3628, line 1: Is there any way we can refer to this file without duplicating it in the repo? Do you need the entire file or can you pass a skeleton or smaller version of it? Same with the passes.json.

src/Analyzer.Cli.UnitTests/CommandLineParserTests.cs
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        [DataTestMethod]
+        [DataRow("Path does not exist", 2, DisplayName = "Invalid template file path provided")]
+        [DataRow("\\CommandLineParserTests.cs", 4, DisplayName = "Path exists, not an ARM template.")]
+        [DataRow("\\CommandLineParserTests.cs", 3, "--report-format", "Sarif", DisplayName = "Path exists, Report-format flag set, --output-file-path flag not included.")]
+        [DataRow("\\CommandLineParserTests.cs", 1, "--parameters-file-path", DisplayName = "Path exists, Parameters-file-path flag included, but no value provided.")]
+        [DataRow("\\AppServicesLogs-Failures.json", 5, DisplayName = "Issues found in the template")]
+        [DataRow("\\AppServicesLogs-Passes.json", 0, DisplayName = "Success")]
+        public void AnalyzeTemplate_ValidInputValues_ReturnExitCode(string relativeTemplatePath, int expectedExitCode, params string[] additionalParams)
```
yane3628, line 30:
```suggestion
        public void AnalyzeTemplate_ValidInputValues_ReturnExpectedExitCode(string relativeTemplatePath, int expectedExitCode, params string[] additionalParams)
```

src/Analyzer.Cli.UnitTests/CommandLineParserTests.cs
```diff
@@ -0,0 +1,98 @@
...
+            args = args.Concat(additionalParams).ToArray();
+            var result = _commandLineParser.InvokeCommandLineAPIAsync(args);
+
+            Assert.AreEqual(expectedExitCode, result.Result);
+        }
+
+        [DataTestMethod]
+        [DataRow("\\Configuration.json", 1, DisplayName = "Provided parameters file is not a parameters file")]
+        [DataRow("\\Parameters.json", 5, DisplayName = "Provided parameters file correct, issues in template")]
+        public void AnalyzeTemplate_ParameterFileParamUsed_ReturnExitCode(string relativeParametersFilePath, int expectedExitCode)
```
yane3628, line 42:
```suggestion
        public void AnalyzeTemplate_ParameterFileParamUsed_ReturnExpectedExitCode(string relativeParametersFilePath, int expectedExitCode)
```

src/Analyzer.Cli.UnitTests/CommandLineParserTests.cs
yane3628, line 53:
```suggestion
        public void AnalyzeTemplate_ConfigurationFileParamUsed_ReturnExpectedExitCode()
```

src/Analyzer.Cli.UnitTests/CommandLineParserTests.cs
```diff
@@ -0,0 +1,98 @@
...
+            var templatePath = String.Concat(Directory.GetCurrentDirectory(), "\\AppServicesLogs-Failures.json");
+            var configurationPath = String.Concat(Directory.GetCurrentDirectory(), "\\Configuration.json");
+            var args = new string[] { "analyze-template", templatePath, "--config-file-path", configurationPath};
+            var result = _commandLineParser.InvokeCommandLineAPIAsync(args);
+
+            Assert.AreEqual(5, result.Result);
+        }
+
+        [TestMethod]
+        public void AnalyzeTemplate_SarifRun_ReturnExitCode()
```
yane3628, line 64:
```suggestion
        public void AnalyzeTemplate_SarifRun_ReturnExpectedExitCode()
```

src/Analyzer.Cli.UnitTests/CommandLineParserTests.cs
```diff
@@ -0,0 +1,98 @@
...
+            
+            File.Delete(outputFilePath);
+        }
+
+        [DataTestMethod]
+        [DataRow(false, 2, DisplayName = "Invalid directory path provided")]
+        [DataRow(true, 3, "--report-format", "Sarif", DisplayName = "Directory exists, Report-format flag set, --output-file-path flag not included.")]
+        [DataRow(true, 1, "--report-format", "Console", "--output-file-path", DisplayName = "Path exists, Report-format flag set, --output-file-path flag included, but no value provided.")]
+        [DataRow(true, 6, DisplayName = "Error + Issue: Scan has both errors and issues")]
+        public void AnalyzeDirectory_ValidInputValues_ReturnExitCode(bool useTestDirectoryPath, int expectedExitCode, params string[] additionalParams)
```
yane3628, line 84:
```suggestion
        public void AnalyzeDirectory_ValidInputValues_ReturnExpectedExitCode(bool useTestDirectoryPath, int expectedExitCode, params string[] additionalParams)
```

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -23,6 +23,16 @@ internal class CommandLineParser
 
         private readonly TemplateAnalyzer templateAnalyzer;
 
+        private enum ExitCode {
```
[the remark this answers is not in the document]
yane3628: I'm not sure how this can be done in this library since I think help is auto-generated. Maybe we can just create an issue for later.

yane3628 (approved):

src/Analyzer.Cli.UnitTests/AppServicesLogs-Failures.json
yane3628: They can be created in ClassInit, then they should only be created once per test class instance.

src/Analyzer.Cli.UnitTests/CommandLineParserTests.cs
yane3628: I think if the sarif output is tested somewhere else it should be OK to not test here. If we want to test that something was actually output to the correct path, we can probably test that part of the functionally if it's not tested in the Reports tests without importing the writer. i.e. If you hard-code the expected sarif output somewhere in the test project and just string compare it to whatever is written to the outputfile. Only thin about this is that if we make updates to the sarif output properties, then this test will have to be updated, but that shouldn't be an issue or a big deal.

JohnathonMohr: Just want to point out that functional tests don't necessarily have to be end-to-end.  Functional will test a larger area than a unit, but it can still be smaller than end to end.  Just depends on what the test goal is.

src/Analyzer.Cli.UnitTests/CommandLineParserTests.cs
```diff
@@ -0,0 +1,98 @@
...
+            Assert.AreEqual(5, result.Result);
+        }
+
+        [TestMethod]
+        public void AnalyzeTemplate_SarifRun_ReturnExitCode()
+        {
+            var templatePath = String.Concat(Directory.GetCurrentDirectory(), "\\AppServicesLogs-Failures.json");
+            var outputFilePath = String.Concat(Directory.GetCurrentDirectory(), "\\outPutFile.sarif");
+            // The FileStream has to be closed for the cli to be able to write values into it
+            using (var fs = File.Create(outputFilePath))
```
JohnathonMohr, line 69: Why do you need to create the file first?

src/Analyzer.Cli.UnitTests/CommandLineParserTests.cs
JohnathonMohr: Same as above - can you create a more descriptive name?

src/Analyzer.Cli.UnitTests/CommandLineParserTests.cs
```diff
@@ -0,0 +1,98 @@
...
+        [DataRow(true, 3, "--report-format", "Sarif", DisplayName = "Directory exists, Report-format flag set, --output-file-path flag not included.")]
+        [DataRow(true, 1, "--report-format", "Console", "--output-file-path", DisplayName = "Path exists, Report-format flag set, --output-file-path flag included, but no value provided.")]
+        [DataRow(true, 6, DisplayName = "Error + Issue: Scan has both errors and issues")]
+        public void AnalyzeDirectory_ValidInputValues_ReturnExpectedExitCode(bool useTestDirectoryPath, int expectedExitCode, params string[] additionalParams)
+        {
+            string[] args;
+            if (useTestDirectoryPath)
+                args = new string[] { "analyze-directory", Directory.GetCurrentDirectory() };
+            else
+                args = new string[] { "analyze-directory", "Directory does not exist" };
```
JohnathonMohr, line 90:
```suggestion
            var args = new string[] { "analyze-directory", useTestDirectoryPath ? Directory.GetCurrentDirectory() : "Directory does not exist" };
```

src/Analyzer.Cli.UnitTests/CommandLineParserTests.cs
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        [DataTestMethod]
+        [DataRow("Path does not exist", 2, DisplayName = "Invalid template file path provided")]
+        [DataRow("\\CommandLineParserTests.cs", 4, DisplayName = "Path exists, not an ARM template.")]
+        [DataRow("\\CommandLineParserTests.cs", 3, "--report-format", "Sarif", DisplayName = "Path exists, Report-format flag set, --output-file-path flag not included.")]
+        [DataRow("\\CommandLineParserTests.cs", 1, "--parameters-file-path", DisplayName = "Path exists, Parameters-file-path flag included, but no value provided.")]
+        [DataRow("\\AppServicesLogs-Failures.json", 5, DisplayName = "Issues found in the template")]
+        [DataRow("\\AppServicesLogs-Passes.json", 0, DisplayName = "Success")]
+        public void AnalyzeTemplate_ValidInputValues_ReturnExpectedExitCode(string relativeTemplatePath, int expectedExitCode, params string[] additionalParams)
```
JohnathonMohr, line 30: Rename `additionalParams` to something more descriptive, like `cliOptions` or `additionalCliOptions`.  Same with other tests.

src/Analyzer.Cli.UnitTests/CommandLineParserTests.cs
```diff
@@ -0,0 +1,98 @@
...
+            var templatePath = String.Concat(Directory.GetCurrentDirectory(), "\\AppServicesLogs-Failures.json");
+            var parametersFilePath = String.Concat(Directory.GetCurrentDirectory(), relativeParametersFilePath);
+            var args = new string[] { "analyze-template", templatePath, "--parameters-file-path", parametersFilePath };
+            var result = _commandLineParser.InvokeCommandLineAPIAsync(args);
+
+            Assert.AreEqual(expectedExitCode, result.Result);
+        }
+
+        [TestMethod]
+        public void AnalyzeTemplate_ConfigurationFileParamUsed_ReturnExpectedExitCode()
```
JohnathonMohr, line 53: It's not clear to me what this scenario is actually testing. Can you create a more descriptive name? (The CLI will always return some kind of exit code. What is special about it when using a configuration file? Will it always be the same when using a configuration file?)  Same with the Sarif test variant below.

JohnathonMohr (changes requested):

src/Analyzer.Cli.UnitTests/CommandLineParserTests.cs
JohnathonMohr: Also @anaismiller you can run the raw CLI and get the output text from the full run with `System.Diagnostics.Process.Start()`, so you don't have to worry about the ReportWriter classes.

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -17,12 +17,26 @@
 
 namespace Microsoft.Azure.Templates.Analyzer.Cli
 {
-    internal class CommandLineParser
+    /// <summary>
+    /// Creates the command line for running the Template Analyzer. 
+    /// Instantiates arguments that can be passed and different commands that can be invoked.
+    /// </summary>
+    public class CommandLineParser
```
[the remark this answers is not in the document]
anaismiller: Oh right! Whoops

VeraBE (approved):

src/Analyzer.Cli/CommandLineParser.cs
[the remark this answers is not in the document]
anaismiller: If you don't mind, can you create the issue please? I don't have a reference for what the help flag looks like or a vision in mind. I'm assuming the help flag is different from the `--verbose` flag?

src/Analyzer.Cli.UnitTests/CommandLineParserTests.cs
anaismiller: Because a file needs to be passed in with `Sarif` flag. I thought it would be weird to pass in an empty file just for this test to work.

src/Analyzer.Cli.UnitTests/AppServicesLogs-Failures.json
anaismiller: Error on my part, I shortened the "Passes" and forgot to do the same with the "Failures"

src/Analyzer.Cli.UnitTests/CommandLineParserTests.cs
anaismiller: Thank you. Yane3628's point about hard-coded expectation and updating the test if a property changes and JohnathonMohr;s point that functional tests != end-to-end made me decide to leave the test as it is.

src/Analyzer.Cli.UnitTests/CommandLineParserTests.cs
JohnathonMohr: The file path with the Sarif flag just tells the CLI where to write the Sarif data to.  It doesn't mean the file needs to exist first.  If it did, then users would have to create an empty file before they could run the command, which would be a poor experience.

src/Analyzer.Cli.UnitTests/AppServicesLogs-Failures.json
anaismiller: I meant instead of declaring these as files, I create a file, add the text to it, then delete is and the end of the test

anaismiller: Deciding I will leave the test as it is.

src/Analyzer.Cli.UnitTests/CommandLineParserTests.cs
anaismiller: Thank you for explaining.

src/Analyzer.Cli/CommandLineParser.cs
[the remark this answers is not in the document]
yane3628: Thinking about it more, maybe it doesn't make sense since the exit codes are not something that users explicitly see anyway.

JohnathonMohr (approved):

src/Analyzer.Cli/ExitCode.cs
```diff
@@ -0,0 +1,46 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Analyzer.Cli
```
yane3628, line 4:
```suggestion
namespace Microsoft.Azure.Templates.Analyzer.Cli
```

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -14,9 +14,14 @@
 using Microsoft.Azure.Templates.Analyzer.Reports;
 using Microsoft.Extensions.Logging;
 using Newtonsoft.Json.Linq;
+using Analyzer.Cli;
```
yane3628, line 17: After the namespace change in ExitCode.cs, this will no longer be needed.

src/Analyzer.Cli/CommandLineParser.cs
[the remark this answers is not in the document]
anaismiller: Okay

yane3628 (approved):

src/Analyzer.Cli/CommandLineParser.cs
JohnathonMohr: Good catch!

PR #240, opened 2022-04-25
Output logs as SARIF tool notifications

### LINKED ISSUE #210: Report error messages in sarif output
Errors should be logged to toolExecutionNotifications property in sarif 
For example, template parsing errors should be logged in the sarif output.

VeraBE:
## Description

Closes #210 

Documentation followed to create the new logger provider: https://docs.microsoft.com/en-us/dotnet/core/extensions/custom-logging-provider

The SARIF spec (https://docs.oasis-open.org/sarif/sarif/v2.1.0/csprd01/sarif-v2.1.0-csprd01.html) says that "The presence within this array of any notification object whose level property (§3.58.6) is "error" SHALL mean that the run failed.", so some cases that previously were logged as errors are now logged as warnings

---

- [ ] Pull request includes test coverage for the included changes.


src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -199,7 +202,18 @@ private int AnalyzeTemplate(FileInfo templateFilePath, FileInfo parametersFilePa
...
+            // Check that output file path provided for sarif report
+            if (reportFormat == ReportFormat.Sarif && outputFilePath == null)
+            {
+                // We can't use the logger for this error,
+                // because we need to get the writer to create the logger, but this check has to be done before getting the writer:
+                Console.WriteLine("Output file path was not provided.");
+                return;
+            }
+
+            var reportWriter = this.GetReportWriter(reportFormat.ToString(), outputFilePath, directoryPath.FullName);
```
JohnathonMohr, line 214: Can you keep the `using` here?

```suggestion
            using var reportWriter = this.GetReportWriter(reportFormat.ToString(), outputFilePath, directoryPath.FullName);
```

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -226,36 +233,37 @@ private void AnalyzeDirectory(DirectoryInfo directoryPath, FileInfo configuratio
...
+                    var errorMessage = $"Unable to analyze {filesFailed.Count} file(s):";
+
+                    foreach (FileInfo failedFile in filesFailed)
                     {
-                        logger.LogError("Unable to analyze {numFilesFailed} file(s):", filesFailed.Count);
-                        foreach (FileInfo failedFile in filesFailed)
-                        {
-                            logger.LogError("\t{failedFile}", failedFile);
-                        }
+                        errorMessage += $" {failedFile}";
```
JohnathonMohr, line 258: Can we comma-separate these instead of just spaces?

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -226,36 +233,37 @@ private void AnalyzeDirectory(DirectoryInfo directoryPath, FileInfo configuratio
...
                     }
+
+                    logger.LogError(errorMessage);
```
JohnathonMohr, line 261:
``` suggestion
                    logger.LogError($"Unable to analyze {filesFailed.Count} file(s): {string.Join(",", filesFailed)}";
```

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -331,7 +324,14 @@ private static ILogger CreateLogger(bool verbose)
                     });
             });
 
+            if (reportFormat == ReportFormat.Sarif && reportWriter != null)
```
JohnathonMohr, line 327: I personally think we shouldn't check `reportWriter` for null here - since it's a private method, it can only be called from within this class, so we should just make sure it always has a value when calling the method (which it does).

src/Analyzer.Cli/SarifNotificationLogger.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// <inheritdoc/>
+        public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
+        {
+            if (!IsEnabled(logLevel))
+            {
+                return;
+            }
+
+            var notificationMessage = state.ToString();
+            if (exception != null)
```
JohnathonMohr, line 48: Outer `if` is not needed

src/Analyzer.Cli/SarifNotificationLogger.cs
```diff
@@ -0,0 +1,68 @@
...
+            {
+                while (exception != null)
+                {
+                    notificationMessage += (" - " + exception.Message + " - " + exception.StackTrace);
+                    exception = exception.InnerException;
+                }
+            }
+
+            var failureLevel = logLevel == LogLevel.Error ? FailureLevel.Error : FailureLevel.Warning;
```
JohnathonMohr, line 57: We can't log info?

src/Analyzer.Cli/SarifNotificationLogger.cs
```diff
@@ -0,0 +1,68 @@
...
+            {
+                return;
+            }
+
+            var notificationMessage = state.ToString();
+            if (exception != null)
+            {
+                while (exception != null)
+                {
+                    notificationMessage += (" - " + exception.Message + " - " + exception.StackTrace);
```
JohnathonMohr, line 52: Instead of putting all the exception data into `Message`, can we use the `Exception` property of `Notification` for this?

```diff
            var notification = new Notification
            {
                Message = new Message { Text = notificationMessage },
                Level = failureLevel,
+               Exception = ExceptionData.Create(exception)
            };
```

src/Analyzer.Cli/SarifNotificationLoggerProvider.cs
```diff
@@ -0,0 +1,39 @@
...
+
+namespace Microsoft.Azure.Templates.Analyzer.Cli
+{
+    /// <summary>
+    /// Class that creates a logger to log warnings and errors as tool notifications in the SARIF output
+    /// </summary>
+    public class SarifNotificationLoggerProvider : ILoggerProvider
+    {
+        /// <summary>
+        /// Class used to output information to the SARIF file
```
JohnathonMohr, line 15:
``` suggestion
        /// Logger used to output information to the SARIF file
```

src/Analyzer.Reports/SarifReportWriter.cs
```diff
@@ -18,13 +18,17 @@ namespace Microsoft.Azure.Templates.Analyzer.Reports
     /// </summary>
     public class SarifReportWriter : IReportWriter
     {
+        /// <summary>
+        /// Class used to output information to the SARIF file
```
JohnathonMohr, line 22:
``` suggestion
        /// Logger used to output information to the SARIF file
```

src/Analyzer.Reports/SarifReportWriter.cs
```diff
@@ -18,13 +18,17 @@ namespace Microsoft.Azure.Templates.Analyzer.Reports
...
+        /// </summary>
+        public SarifLogger sarifLogger;
```
JohnathonMohr, line 24: Make this a property instead of a variable.  It's generally best practice to have public members be properties instead of just variables.  We should include this in the contribution guide at some point.
There's not really an ideal place for properties in this class right now - I think I would move it to just below the constructor.

```cs
        public SarifLogger SarifLogger { get; set; }
```

JohnathonMohr (changes requested):

src/Analyzer.Cli/CommandLineParser.cs
JohnathonMohr: This doesn't really display very well.  Some of the lines shown here as removed were already removed as part of your PR, so this isn't an accurate representation of the real proposed change.

src/Analyzer.Cli/SarifNotificationLogger.cs
VeraBE: I didn't see the Create function before, and was having a hard time using the constructor of ExceptionData

src/Analyzer.Cli/SarifNotificationLogger.cs
VeraBE: Yes, I don't know why I assumed we only wanted to log warnings and errors there

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -287,37 +288,20 @@ private bool IsValidSchema(string template)
...
-            {
-                exception = exception.InnerException;
-                exceptionMessage += getExceptionInfo(exception);
-            }
-
-            return exceptionMessage;
-        }
-
-        private IReportWriter GetReportWriter(string reportFormat, FileInfo outputFile, string rootFolder = null)
+        private IReportWriter GetReportWriter(ReportFormat reportFormat, FileInfo outputFile, string rootFolder = null)
```
yane3628, line 291:
```suggestion
        private static IReportWriter GetReportWriter(ReportFormat reportFormat, FileInfo outputFile, string rootFolder = null)
```

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -226,36 +233,30 @@ private void AnalyzeDirectory(DirectoryInfo directoryPath, FileInfo configuratio
...
+                if (filesFailed.Count > 0)
+                {
+                    logger.LogError($"Unable to analyze {filesFailed.Count} file(s): {string.Join(", ", filesFailed)}");
+                }
             }
-            catch (Exception exp)
+            catch (Exception exception)
             {
-                logger.LogError(GetExceptionMessage(exp));
+                logger.LogError(exception, "An exception occurred while analyzing the directory provided");
```
yane3628, line 259: Why did you remove reporting of the inner exception?

yane3628 (commented):
If BPA is invoked directly through the library instead of the CLI, will the SARIF still have the tool notifications of exceptions of errors when processing and evaluating the template?

yane3628:
Once https://github.com/Azure/template-analyzer/pull/239 is complete, it would be nice to add some tests for this new feature.

src/Analyzer.Cli/CommandLineParser.cs
yane3628: What about in the console output scenario? Or in future output type scenarios? Will each output type have to account for iterating through the inner exceptions?

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -131,11 +131,27 @@ private RootCommand SetupCommandLineAPI()
...
+                Console.WriteLine("Output file path was not provided.");
+                return 3;
             }
 
-            bool disposeWriter = false;
+            if (writer == null)
+            {
+                writer = GetReportWriter(reportFormat, outputFilePath);
+                disposeWriter = true;
+            }
```
JohnathonMohr, line 149: You could combine this with the check above if you wanted.

```
if (writer == null)
{
    // if report and filePath aren't valid, return error.
    // otherwise create writer and set disposal.
}
```

src/Analyzer.Cli/CommandLineParser.cs
VeraBE: For the console output it's handled by the general logger library, and for the SARIF output is handled by the SARIF logger library

VeraBE:
> If BPA is invoked directly through the library instead of the CLI, will the SARIF still have the tool notifications of exceptions of errors when processing and evaluating the template?

@yane3628 I don't see the SarifReportWriter being used anywhere else besides the CLI, I don't think we produce a Sarif file in the library

JohnathonMohr (approved):

src/Analyzer.Cli/CommandLineParser.cs
VeraBE: @yane3628 Is it okay if I add an issue to improve the formatting of the console output? I think it's out of scope for this PR

yane3628: Yea, I think making an issue to solve later is OK.

yane3628:
> > If BPA is invoked directly through the library instead of the CLI, will the SARIF still have the tool notifications of exceptions of errors when processing and evaluating the template?
> 
> @yane3628 I don't see the SarifReportWriter being used anywhere else besides the CLI, I don't think we produce a Sarif file in the library

Mmm, this makes sense. If the library is invoked directly, then the results will be in the Evaluation object. Are the exceptions/errors something we're OK "losing" in certain scenarios since we're relying on the caller of the library to pass a logger? This definitely is something out of scope of this PR, but something worth thinking about.

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -131,11 +131,26 @@ private RootCommand SetupCommandLineAPI()
...
         { 
-            if (logger == null) {
-                logger = CreateLogger(verbose);
+            bool disposeWriter = false;
+
+            if (writer == null)
+            {
+                if (reportFormat == ReportFormat.Sarif && outputFilePath == null)
+                {
+                    // We can't use the logger for this error,
```
anaismiller, line 140: improve comment line spacing

src/Analyzer.Cli.FunctionalTests/Analyzer.Cli.FunctionalTests.csproj
```diff
@@ -33,4 +33,9 @@
     </None>
   </ItemGroup>
 
-</Project>
+  <Target Name="CopyTestFolder" AfterTargets="Build" Condition="!Exists('$(OutputPath)\ADirectoryToAnalyze')">
+    <Exec Command="cp -r ../Analyzer.Cli.FunctionalTests/ADirectoryToAnalyze $(OutputPath)/ADirectoryToAnalyze/" Condition="'$(OS)' == 'Unix'" />
+    <Exec Command="Xcopy /E /I ..\Analyzer.Cli.FunctionalTests\ADirectoryToAnalyze $(OutputPath)\ADirectoryToAnalyze\" Condition="'$(OS)' != 'Unix'" />
+  </Target>
```
JohnathonMohr, line 39: You shouldn't need this.  Just add additional `<None Update ... >`s above.  You can even do `ADirectoryToAnalyze\*` to output all of them in the directory.

src/Analyzer.Cli.FunctionalTests/CommandLineParserTests.cs
```diff
@@ -86,5 +87,35 @@ public void AnalyzeDirectory_ValidInputValues_ReturnExpectedExitCode(bool useTes
 
             Assert.AreEqual(expectedExitCode, result.Result);
         }
+
+        [TestMethod]
+        public void AnalyzeDirectory_DirectoryWithOtherJsonFiles_LogsExpectedErrorInSarif()
+        {
+            var outputFilePath = Path.Combine(Directory.GetCurrentDirectory(), "Output.json");
```
JohnathonMohr, line 94:
```suggestion
            var outputFilePath = Path.Combine(Directory.GetCurrentDirectory(), "Output.sarif");
```

src/Analyzer.PowerShellRuleEngine/PowerShellRuleEngine.cs
```diff
@@ -79,7 +79,7 @@ public IEnumerable<IEvaluation> AnalyzeTemplate(TemplateContext templateContext,
             {
                 // There was an error loading the TTK module.  Return an empty collection.
 
-                logger?.LogError("There was an error running the PowerShell based checks");
+                logger?.LogWarning("There was an error running the PowerShell based checks");
```
yane3628, line 82: Do we want to change the text to not include the word error any more since it's now of type warning?

src/Analyzer.TemplateProcessor/ArmTemplateProcessor.cs
```diff
@@ -141,7 +141,7 @@ internal Template ParseAndValidateTemplate(InsensitiveDictionary<JToken> paramet
 
                 // Do not throw if there was another issue with evaluating language expressions
 
-                logger?.LogError(ex, "An exception occurred when processing the template language expressions");
+                logger?.LogWarning(ex, "An exception occurred when processing the template language expressions");
```
yane3628, line 144: Same with the changes to warning on this file.

yane3628 (approved):

src/Analyzer.PowerShellRuleEngine/PowerShellRuleEngine.cs
VeraBE: I'm inclined to leave the messages as they are, because there was an error running PowerShell, is just that we might not consider that a total failure of the tool

PR #245, opened 2022-06-01
CLI refactoring and other improvements

### LINKED ISSUE #207: Extract common functionality between AnalyzeDirectory and AnalyzeTemplate in CLI

### LINKED ISSUE #233: Move logic of reading configuration file to cli and add positive tests for function [#56441]

Currently, the `Analyzer.Cli.CommandLineParser` calls `Analyzer.Core.FilterRules` to read the configuration file and call `jsonRuleEngine.FilterRules` to filter the rules.
Instead, the logic for reading the configuration file should be moved to `Analyzer.Cli.CommandLineParser` project (entire `GetConfigurationFileContents` function) and only the read file should be passed to `Analyzer.Core.FilterRules` that will then call `jsonRuleEngine.FilterRules` to filter the rules. The tests should also be moved to this project. This needs #28 for the test project.

In addition, positive tests need to be added to test that `jsonRuleEngine.FilterRules` is using the configuration file. This can also result in functional tests to verify the outcome of the function returns the filtered rules.

### Additional context

Mocking the class can be helpful. The JsonRuleEngine or IRuleEngine can be used for mocking. This was not done when Severities was first introduced as mocking for positive tests involved a lot more work than was feasible and reasonable. Note that this may involve adding additional projects to the test project just for testing the using of the configuration file.

JohnathonMohr:
## Description
Refactor CLI
- Extract commonalities between template and directory analysis into setup and cleanup functions (closes #207).
- Move parsing of configuration to CLI (closes #233).
- Improved directory searching for template files - use .NET platform to filter for JSON files
- [NEW] Analyze-Template no longer requires the file to have a ".json" file extension; only the JSON properties will be validated.

Other improvements
- Update template validation to only read what's necessary of a file, instead of reading the whole file into memory first.
- Polish summary output
  - Update text for counts
  - Fix counts in some scenarios
  - Remove "Unable to analyze n files" from summary, as that line is already a summary by itself.

---

### DIFF
--- a/src/Analyzer.Cli/CommandLineParser.cs
+++ b/src/Analyzer.Cli/CommandLineParser.cs
@@ -185,7 +185,7 @@ namespace Microsoft.Azure.Templates.Analyzer.Cli
                 return (int)ExitCode.ErrorInvalidARMTemplate;
             }
 
-            var analysisResult = AnalyzeTemplate(templateFilePath, parametersFilePath, this.reportWriter, this.logger);
+            var analysisResult = AnalyzeTemplate(templateFilePath, parametersFilePath);
 
             FinishAnalysis();
             return (int)analysisResult;
@@ -222,7 +222,7 @@ namespace Microsoft.Azure.Templates.Analyzer.Cli
             var filesFailed = new List<FileInfo>();
             foreach (FileInfo file in filesToAnalyze)
             {
-                ExitCode res = AnalyzeTemplate(file, null, reportWriter, logger);
+                ExitCode res = AnalyzeTemplate(file, null);
 
                 if (res == ExitCode.Success || res == ExitCode.Violation)
                 {
@@ -254,16 +254,16 @@ namespace Microsoft.Azure.Templates.Analyzer.Cli
             return (int)exitCode;
         }
 
-        private ExitCode AnalyzeTemplate(FileInfo templateFilePath, FileInfo parametersFilePath, IReportWriter writer, ILogger logger)
+        private ExitCode AnalyzeTemplate(FileInfo templateFilePath, FileInfo parametersFilePath)
         { 
             try
             {
                 string templateFileContents = File.ReadAllText(templateFilePath.FullName);
                 string parameterFileContents = parametersFilePath == null ? null : File.ReadAllText(parametersFilePath.FullName);
 
-                IEnumerable<IEvaluation> evaluations = this.templateAnalyzer.AnalyzeTemplate(templateFileContents, parameterFileContents, templateFilePath.FullName, logger);
+                IEnumerable<IEvaluation> evaluations = this.templateAnalyzer.AnalyzeTemplate(templateFileContents, parameterFileContents, templateFilePath.FullName);
 
-                writer.WriteResults(evaluations, (FileInfoBase)templateFilePath, (FileInfoBase)parametersFilePath);
+                this.reportWriter.WriteResults(evaluations, (FileInfoBase)templateFilePath, (FileInfoBase)parametersFilePath);
 
                 return evaluations.Any(e => !e.Passed) ? ExitCode.Violation : ExitCode.Success;
             }
@@ -292,7 +292,7 @@ namespace Microsoft.Azure.Templates.Analyzer.Cli
             this.reportWriter = GetReportWriter(reportFormat, outputFilePath, directoryToAnalyze?.FullName);
             CreateLoggers(verbose);
 
-            this.templateAnalyzer = TemplateAnalyzer.Create(runPowershell);
+            this.templateAnalyzer = TemplateAnalyzer.Create(runPowershell, this.logger);
 
             if (!TryReadConfigurationFile(configurationFile, out var config))
             {
--- a/src/Analyzer.Cli/CommandLineParser.cs
+++ b/src/Analyzer.Cli/CommandLineParser.cs
@@ -23,10 +23,12 @@ namespace Microsoft.Azure.Templates.Analyzer.Cli
     /// </summary>
     internal class CommandLineParser
     {
-        RootCommand rootCommand;
+        private RootCommand rootCommand;
+        private TemplateAnalyzer templateAnalyzer;
 
-        private readonly TemplateAnalyzer templateAnalyzer;
-        private readonly SummaryLoggerProvider summaryLoggerProvider;
+        private IReportWriter reportWriter;
+        private ILogger logger;
+        private SummaryLogger summaryLogger;
 
         /// <summary>
         /// Constructor for the command line parser. Sets up the command line API. 
@@ -34,8 +36,6 @@ namespace Microsoft.Azure.Templates.Analyzer.Cli
         public CommandLineParser()
         {
             SetupCommandLineAPI();
-            templateAnalyzer = TemplateAnalyzer.Create();
-            summaryLoggerProvider = new SummaryLoggerProvider();
         }
 
         /// <summary>
@@ -102,7 +102,7 @@ namespace Microsoft.Azure.Templates.Analyzer.Cli
 
             analyzeTemplateCommand.Handler = CommandHandler.Create<FileInfo, FileInfo, FileInfo, ReportFormat, FileInfo, bool, bool>(
                 (templateFilePath, parametersFilePath, configFilePath, reportFormat, outputFilePath, runTtk, verbose) =>
-                this.AnalyzeTemplate(templateFilePath, parametersFilePath, configFilePath, reportFormat, outputFilePath, runTtk, verbose));
+                this.AnalyzeTemplateCommandHandler(templateFilePath, parametersFilePath, configFilePath, reportFormat, outputFilePath, runTtk, verbose));
 
             // Setup analyze-directory w/ directory argument and configuration file option
             Command analyzeDirectoryCommand = new Command(
@@ -126,7 +126,7 @@ namespace Microsoft.Azure.Templates.Analyzer.Cli
 
             analyzeDirectoryCommand.Handler = CommandHandler.Create<DirectoryInfo, FileInfo, ReportFormat, FileInfo, bool, bool>(
                 (directoryPath, configurationsFilePath, reportFormat, outputFilePath, runTtk, verbose) =>
-                this.AnalyzeDirectory(directoryPath, configurationsFilePath,  reportFormat, outputFilePath, runTtk, verbose));
+                this.AnalyzeDirectoryCommandHandler(directoryPath, configurationsFilePath,  reportFormat, outputFilePath, runTtk, verbose));
 
             // Add commands to root command
             rootCommand.AddCommand(analyzeTemplateCommand);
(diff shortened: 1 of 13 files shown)


src/Analyzer.Cli/Loggers/SummaryLogger.cs
```diff
@@ -56,42 +65,48 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
         /// <summary>
         /// Outputs to console a summary of the errors and warnings logged
         /// </summary>
-        public void SummarizeLogs(bool showVerboseMessage)
+        public void SummarizeLogs()
         {
+            // Local function to describe a count by returning the number with the appropriate form
+            // (singular or plural - just adding 's') of its description
+            string DescNum(int count, string description) => $"{count} {description}{(count == 1 ? "" : "s")}";
```
anaismiller, line 72: Can you rename this func to PluralizeDescrip?

src/Analyzer.JsonRuleEngine/JsonRuleEngine.cs
```diff
@@ -52,47 +52,37 @@ public static JsonRuleEngine Create(string rawRuleDefinitions, BuildILineNumberR
...
-        /// Modifies the rules to run based on values defined in the configurations file.
+        /// Modifies the rules to run based on values defined in the configuration file.
         /// </summary>
         /// <param name="configuration">The configuration specifying rule modifications.</param>
-        public void FilterRules(string configuration)
+        public void FilterRules(ConfigurationDefinition configuration)
         {
-            if (string.IsNullOrEmpty(configuration))
-                return;
+            if (configuration == null)
```
anaismiller, line 60: A configuration is an optional parameter

src/Analyzer.Reports/SarifReportWriter.cs
```diff
@@ -206,6 +209,8 @@ public void Dispose(bool disposing)
             this.SarifLogger?.Dispose();
             this.outputTextWriter?.Dispose();
             this.reportFileStream?.Dispose();
+
+            Console.WriteLine($"{Environment.NewLine}Wrote {totalResults} results to {reportFile.FullName}");
```
anaismiller, line 213: What about the singular form of result for 1 item?

src/Analyzer.Core/TemplateAnalyzer.cs
```diff
@@ -21,21 +21,25 @@ namespace Microsoft.Azure.Templates.Analyzer.Core
...
             this.jsonRuleEngine = jsonRuleEngine;
+            this.powerShellRuleEngine = powerShellRuleEngine;
         }
 
         /// <summary>
         /// Creates a new <see cref="TemplateAnalyzer"/> instance with the default built-in rules.
         /// </summary>
-        public static TemplateAnalyzer Create()
+        /// <param name="usePowerShell">Whether or not to use PowerShell rules to analyze the template.</param>
+        public static TemplateAnalyzer Create(bool usePowerShell)
```
anaismiller, line 42: Is it worth putting a default value? That way in the tests you don't have to put "false"

src/Analyzer.Cli/Loggers/SummaryLogger.cs
JohnathonMohr: I almost just named it "Pluralize", but I decided it's a little bit of a misnomer, since it won't actually pluralize if the count is 1.  I would be fine with "Pluralize" if that's preferable to "Describe Number".

src/Analyzer.JsonRuleEngine/JsonRuleEngine.cs
JohnathonMohr: The configuration itself is optional.  But if the `FilterRules` method is called, it shouldn't be null.  (If there is no configuration, this function shouldn't be called.)

src/Analyzer.Cli/Loggers/SummaryLogger.cs
anaismiller: I'm not a fan of "Describe Number." It confused me. I will let others share their input.

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -154,171 +153,176 @@ private RootCommand SetupCommandLineAPI()
...
         {
-            foreach (FileInfo file in directoryPath.GetFiles())
-            {
-                if (file.Extension.Equals(".json", StringComparison.OrdinalIgnoreCase))
+            this.reportWriter?.Dispose();
+        }
+
+        private IEnumerable<FileInfo> FindTemplateFilesInDirectory(DirectoryInfo directoryPath) =>
+            directoryPath.GetFiles(
+                "*.json",
```
anaismiller, line 314: In the PR, you said it does not need to be a ".json" ending. Am I understanding this correctly that you are getting-the-files with a ".json" ending only?

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -350,42 +354,74 @@ private bool IsValidTemplate(string template)
...
-                case ReportFormat.Sarif:
-                    return new SarifReportWriter((FileInfoBase)outputFile, rootFolder);
-                case ReportFormat.Console:
-                    return new ConsoleReportWriter();
-                default:
-                    return new ConsoleReportWriter();
-            }
-        }
+        private static IReportWriter GetReportWriter(ReportFormat reportFormat, FileInfo outputFile, string rootFolder = null) =>
+            reportFormat switch {
```
anaismiller, line 358: Why make this a switch and not an if/else? I think prior it was to make each case be explicitly defined or for the scenario where more cases would be added

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -350,42 +354,74 @@ private bool IsValidTemplate(string template)
...
             });
 
-            if (reportFormat == ReportFormat.Sarif)
+            if (this.reportWriter is SarifReportWriter sarifWriter)
             {
-                var sarifLogger = ((SarifReportWriter)reportWriter).SarifLogger;
+                loggerFactory.AddProvider(new SarifNotificationLoggerProvider(sarifWriter.SarifLogger));
+            }
+
+            this.logger = loggerFactory.CreateLogger("TemplateAnalyzerCLI");
```
anaismiller, line 383: Not from your PR, but should the "LI" be "li" lowercased here?

src/Analyzer.Core/TemplateAnalyzer.cs
```diff
@@ -114,47 +116,12 @@ private static string LoadRules()
...
-            {
-                string path;
-                if (configurationsFilePath != null)
-                {
-                    path = configurationsFilePath.FullName;
-                }
-                else
-                {
-                    path = Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location),
-                        "Configurations", "Configuration.json");
```
anaismiller, line 133: Where is this implemented? I don't see it in Cli.CommandLineParser.ReadConfigurationFile()

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -154,171 +153,176 @@ private RootCommand SetupCommandLineAPI()
...
-                    logger.LogError($"Unable to analyze {filesFailed.Count} file(s): {string.Join(", ", filesFailed)}");
-                    return (int)(issueReported ? ExitCode.ErrorAndViolation : ExitCode.ErrorGeneric);
-                }
-                return issueReported ? (int)ExitCode.Violation : (int)ExitCode.Success;
-            }
-            catch (Exception exception)
-            {
-                logger.LogError(exception, "An exception occurred while analyzing the directory provided");
-                return (int)ExitCode.ErrorGeneric;
-            }
```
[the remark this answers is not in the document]
JohnathonMohr: I moved a lot of code around, so the diff here isn't great.  This used to be in AnalyzeDirectory.  It's no longer in AnalyzeDirectoryCommandHandler because nothing there is expected to throw an exception (AnalyzeTemplate will catch anything itself and return gracefully).

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -154,171 +153,176 @@ private RootCommand SetupCommandLineAPI()
...
-                }
-
-                templateAnalyzer.FilterRules(configurationsFilePath);
-
-                // Find files to analyze
-                var filesToAnalyze = new List<FileInfo>();
-                FindJsonFilesInDirectoryRecursive(directoryPath, filesToAnalyze);
+                ConfigurationDefinition config = ReadConfigurationFile(configurationFilePath);
+                if (config == null)
+                    return ExitCode.ErrorGeneric;
```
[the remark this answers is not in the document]
JohnathonMohr: I considered it.  It would be "out of order", separated from other related exit codes, with the way they've been defined so far.

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -154,171 +153,176 @@ private RootCommand SetupCommandLineAPI()
...
 
-            using var reportWriter = GetReportWriter(reportFormat, outputFilePath, directoryPath.FullName);
+            this.reportWriter = GetReportWriter(reportFormat, outputFilePath, directoryToAnalyze?.FullName);
+            CreateLoggers(verbose);
 
-            var logger = CreateLogger(verbose, reportFormat, reportWriter);
+            this.templateAnalyzer = TemplateAnalyzer.Create(runPowershell);
 
-            try
+            if (configurationFilePath != null)
```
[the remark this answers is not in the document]
JohnathonMohr: Oh, it didn't occur to me before your comment that a "default" config would be loaded if it existed without even specifying the argument on the command line!  I'll update that.

src/Analyzer.Cli/CommandLineParser.cs
JohnathonMohr: Analyze-template doesn't require a .json extension. This function is only called for analyze-directory.

src/Analyzer.Core/TemplateAnalyzer.cs
JohnathonMohr: I think I'll leave this without a default parameter.

src/Analyzer.Cli/CommandLineParser.cs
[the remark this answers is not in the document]
anaismiller: (AnalyzeTemplate will catch anything itself and return gracefully) <-- ok, sounds good.

src/Analyzer.Cli/CommandLineParser.cs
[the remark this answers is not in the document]
anaismiller: Being more specific would be better/can help people debugging make it faster. I should have factored the order in to the exitCodes in reverse order.

src/Analyzer.Cli/CommandLineParser.cs
anaismiller: I see, that is why you specifically specified "Analyze-Template". I am sorry about that. 
Follow up, why only Analyze-Template and not Analyze-Directory too?
I know the size is an issue and can impact the runtime, but want to know more. Also on that note, should any Readme files be updated with this addition/change?

src/Analyzer.Cli/CommandLineParser.cs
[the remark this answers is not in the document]
JohnathonMohr: We haven't made a release that includes the exit codes yet - we could reorder them now as an option.

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -292,12 +294,15 @@ private ExitCode AnalyzeTemplate(FileInfo templateFilePath, FileInfo parametersF
...
+            if (!TryReadConfigurationFile(configurationFile, out var config))
             {
-                ConfigurationDefinition config = ReadConfigurationFile(configurationFilePath);
-                if (config == null)
-                    return ExitCode.ErrorGeneric;
+                return ExitCode.ErrorInvalidConfiguration;
+            }
 
+            // Success from TryReadConfigurationFile means there wasn't an error looking for the config.
+            // config could still be null if no path was specified in the command and no default exists.
```
anaismiller, line 303:
```suggestion
            // Config could still be null if no path was specified in the command and no default exists.
```

src/Analyzer.Cli/CommandLineParser.cs
JohnathonMohr: Thanks for the README reminder - there was a separate change I wanted to include that I missed.  But for the .json change, we don't specify a requirement for that extension when scanning a template, so I don't think there's a reason for a change.  I might add something in the directory scan that explains it looks for .json files though.

.json is the standard ARM template extension, so if BPA is performing a directory search on its own, there's no reason to look at every single file to see if it happens to be a template; it should just stick to the standard.  On the other hand, if BPA is invoked to scan a *specific file* on the CLI (and there would only be a single file every time), my thought is there's no reason to *not* try to scan it.

anaismiller (approved):

src/Analyzer.Cli/CommandLineParser.cs
anaismiller: Thumbs-up to first paragraph.

2nd paragraph: It is fine now, sure, but like if the ARM template file is a .txt file instead of a .json file, someone using the tool could get the wrong impression when they use our tool that their .txt file is great. Then when they go to deploy it, the .txt file won't work and they could be confused.

JohnathonMohr: It depends on what you mean by "when they use our tool that their .txt file is great".  Analyze-template will still verify the file is actually a valid ARM template, so if the file is just plain text, it will still return an error.  It just won't require the file extension.

VeraBE: I agree with Anais, if deploying a template using a .txt file returns an error I think we should too

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -154,171 +155,179 @@ private RootCommand SetupCommandLineAPI()
             return rootCommand;
         }
 
-        private int AnalyzeTemplate(FileInfo templateFilePath, FileInfo parametersFilePath, FileInfo configurationsFilePath, ReportFormat reportFormat, FileInfo outputFilePath, bool runTtk, bool verbose, bool printMessageIfNotTemplate = true, IReportWriter writer = null, bool readConfigurationFile = true, ILogger logger = null)
+        private int AnalyzeTemplateCommandHandler(
```
VeraBE, line 158: SummarizeLogs now is only called in the directory handler, I think logs should still be summarized for individual templates also

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -24,37 +23,39 @@ namespace Microsoft.Azure.Templates.Analyzer.Cli
...
-                 "resources",
-                 "outputs",
-        };
-
         /// <summary>
         /// Constructor for the command line parser. Sets up the command line API. 
         /// </summary>
         public CommandLineParser()
         {
             SetupCommandLineAPI();
```
VeraBE, line 58: I think it would be a good idea to also refactor `SetupCommandLineAPI()` to do something like add all the new options to an array, that at the end of the function is used to add the options to both commands, so we don't have to remember to add new options into both every time

src/Analyzer.Cli/CommandLineParser.cs
```diff
@@ -154,171 +155,179 @@ private RootCommand SetupCommandLineAPI()
...
-                // but this check has to be done before getting the writer:
-                Console.WriteLine("When using --report-format sarif flag, --output-file-path flag is required.");
-                return (int)ExitCode.ErrorMissingPath;
+                Console.Error.WriteLine("When using --report-format sarif flag, --output-file-path flag is required.");
+                return ExitCode.ErrorMissingPath;
             }
 
-            using var reportWriter = GetReportWriter(reportFormat, outputFilePath, directoryPath.FullName);
+            this.reportWriter = GetReportWriter(reportFormat, outputFilePath, directoryToAnalyze?.FullName);
+            CreateLoggers(verbose);
```
VeraBE, line 293: Why is it plural now when it's only creating one logger?

src/Analyzer.Cli/CommandLineParser.cs
JohnathonMohr: I just confirmed Azure Powershell will deploy a template file that has a ".txt" extension (it used a .txt parameters file too).  Can I leave this as-is and not require a .json extension?

src/Analyzer.Cli/CommandLineParser.cs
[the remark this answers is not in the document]
JohnathonMohr: We could instead to bitwise operations to create the exit codes.
Example:
1 = rule violation
2 = exception parsing template
4 = error in path specification
8 = error in config
8 = ...
Exit code = "everything that applies, bitwise &'d together"

src/Analyzer.Cli/CommandLineParser.cs
JohnathonMohr: The way I understood the summary logger was to make sure all the errors and warnings that occurred while analyzing several templates in a directory would be clearly presented after all analysis was over.  In the case of a single template, those errors/warnings would just be immediately repeated.  I didn't see any value in doing this, and felt it would be a little annoying instead.

src/Analyzer.Cli/CommandLineParser.cs
JohnathonMohr: I considered that too, but decided against it for some reason.  Since you also support this, I'll include this. 🙂

src/Analyzer.Cli/CommandLineParser.cs
JohnathonMohr: I think of it as creating multiple:
1. Console logger
2. Summary logger
3. Maybe: Sarif logger

src/Analyzer.Cli/CommandLineParser.cs
VeraBE: I guess it depends on the template size/number of results. I can see a single template generating a console output big enough for the summary to make sense anyways

JohnathonMohr: Keeping summary with template analysis.

src/Analyzer.Cli/CommandLineParser.cs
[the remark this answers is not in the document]
JohnathonMohr: I've rearranged the exit codes.  @VeraBE and @anaismiller please take a look.  If it seems ok, I'll update the README.

VeraBE (approved):

src/Analyzer.Cli/ExitCode.cs
```diff
@@ -14,9 +14,10 @@ public enum ExitCode
         Success = 0,
 
         /// <summary>
-        /// Error: GenericError
+        /// Error: There was an error with the command.
+        /// This is returned by rootCommand.InvokeAsync in the CLI if there is an error.
         /// </summary>
-        ErrorGeneric = 1,
+        ErrorCommand = 1,
```
anaismiller, line 20: Please update the ExitCodes in /Readme.md (line ~68)

JohnathonMohr: Yes, that was my plan. :) (https://github.com/Azure/template-analyzer/pull/245#discussion_r893973699)

PR #250, opened 2022-06-14
resourceType now matches expected resources, regardless of how child resources are defined

### LINKED ISSUE #115: Resolve resource type should account for the possibility of the resource being a child resource
When evaluating a rule, resolve resource type should account for the possibility of the resource being a child resource.

For a rule that specifies `"resourceType": "Microsoft.Web/sites/siteextensions"`, resolve resource type should check for child resources in resources of type "Microsoft.Web/sites".  Because the "Microsoft.Web/sites/siteextensions" child resource will have its type specified as simply "siteextensions", resource type selection needs to be intelligent enough to understand the full type of the template resources when searching for the type specified in the rule.

Possible challenge is differentiating between a resource defined in resources vs it's copied value as a child resource if it depends on another resource (see #71).

VeraBE:
## Description

Closes #115 

When processing the template we now replace all the resource types for the full type value. Doing this, if a child resource is defined inside the resources array of the parent, or by itself linked to the parent by a `dependsOn`, we will evaluate the rules on it properly

The resourceType value in rules now should be a full resource type. We only had one rule that didn't use this format, and after these changes, it was possible to simplify the rule definition

Documentation on how child resource types should be written in templates: https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/child-resource-name-type I updated some tests to follow it

The updates in ArmTemplateProcessorTests related to the DeepEquals comparisons are not ideal, I couldn't find a better way

Also, ResolveResourceType now considers the possibility of the searched resource type being a child of one if its prefixes

---


anaismiller (approved):

src/Analyzer.Core.BuiltInRuleTests/Tests/TA-000001/TA-000001.json
```diff
@@ -86,6 +102,10 @@
       {
         "ResourceName": "sitesConfigDependingOnLinuxKind/diagLogsDisabled",
         "Explanation": "The site config configures a site of the wrong kind"
+      },
+      {
+        "ResourceName": "withConfigAsChildResource",
+        "Explanation": "The config child resource has the desired values"
```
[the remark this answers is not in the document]
VeraBE: Yes, it says that the resource of type "config" is a child resource, and it has the desired values

src/Analyzer.Core.BuiltInRuleTests/Tests/TA-000021/AutomationAccountsVariables.json
```diff
@@ -35,6 +35,22 @@
       "name": "automationAccounts/isEncryptedNotDefined",
       "location": "[parameters('location')]",
       "properties": {}
+    },
+    {
+      "type": "Microsoft.Automation/automationAccounts",
+      "apiVersion": "2021-04-01",
+      "name": "justAutomationAccounts",
+      "location": "[parameters('location')]",
+      "resources": [
```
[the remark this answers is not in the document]
VeraBE: The resource of type "variables" was created in the resources array of justAutomationAccounts, so it's its child

src/Analyzer.Utilities/JsonPathResolver.cs
```diff
@@ -71,23 +72,55 @@ public IEnumerable<IJsonPathResolver> Resolve(string jsonPath)
...
+                    if (string.Equals(resource.Value.InsensitiveToken("type")?.Value<string>(), resourceTypePrefixes[numOfPrefix], StringComparison.OrdinalIgnoreCase))
+                    {
+                        if (String.IsNullOrEmpty(resourceTypeSuffixes[numOfPrefix]))
+                        {
+                            yield return new JsonPathResolver(resource.Value, resource.Value.Path, this.resolvedPaths);
+                        }
+                        else
+                        {
+                            // In this case we still haven't matched the suffix of the prefix found
+                            foreach(var newJsonPathResolver in ResolveResourceType(String.Concat(resourceTypePrefixes[numOfPrefix], resourceTypeSeparator, resourceTypeSuffixes[numOfPrefix]), resource.Value.Path, resource.Value))
```
JohnathonMohr, line 116: The `ResolveResourceType` public method shouldn't expose the details of the internal recursion (by having two additional parameters that are only used internally).

An easy way around this, without having to refactor the method at all, is to create a new resolver with a new scope (as was done previously) and call `ResolveResourceType` on that, implicitly using the new scope.
```cs
var subScope = new JsonPathResolver(resource.Value, resource.Value.Path, this.resolvedPaths);
var subResourceType = String.Concat(resourceTypePrefixes[numOfPrefix], resourceTypeSeparator, resourceTypeSuffixes[numOfPrefix]);
foreach(var newJsonPathResolver in subScope.ResolveResourceType(subResourceType))
{
...
}
```

src/Analyzer.TemplateProcessor/ArmTemplateProcessor.cs
```diff
@@ -174,6 +174,8 @@ internal Template ProcessResourcesAndOutputs(Template template)
                 {
                     AddResourceMapping(resourceInfo.expandedPath, resourceInfo.resource.Path);
                 }
+
+                resourceInfo.resource.Type.Value = resourceNameAndType.Split(" ")[1];
```
JohnathonMohr, line 178: This looks like it replaces the type of the child resource with the full type definition that would be required of a top-level resource.

If that's the case, why is there a bunch of string manipulation looking for the short form of a resource type in `JsonPathResolver`?  Shouldn't we expect to always be trying to resolve a fully-typed string, and be comparing it against fully-typed strings?

VeraBE: The changes on that file were all related to the issue I didn't fix in the original version of the PR

"For a rule that specifies "resourceType": "Microsoft.Web/sites/siteextensions", resolve resource type should check for child resources in resources of type "Microsoft.Web/sites". Because the "Microsoft.Web/sites/siteextensions" child resource will have its type specified as simply "siteextensions", resource type selection needs to be intelligent enough to understand the full type of the template resources when searching for the type specified in the rule."

Even if the type siteextensions now would be Microsoft.Web/sites/siteextensions, we still need to look through Microsoft.Web/sites first

src/Analyzer.Utilities/JsonPathResolver.cs
```diff
@@ -71,23 +72,55 @@ public IEnumerable<IJsonPathResolver> Resolve(string jsonPath)
...
 
+            // When trying to resolve Microsoft.Web/sites/siteextensions, for example, we should consider that siteextensions can be a child resource of Microsoft.Web/sites (a prefix of the original resource type)
+            var resourceTypePrefixes = new List<string> { resourceType };
+            var resourceTypeSuffixes = new List<string> { "" };
+            var indexesOfTypesSeparators = Regex.Matches(resourceType, resourceTypeSeparator).Cast<Match>().Select(m => m.Index).Skip(1);
+            foreach (var indexOfTypeSeparator in indexesOfTypesSeparators)
+            {
+                resourceTypePrefixes.Add(resourceType[..indexOfTypeSeparator]);
+                resourceTypeSuffixes.Add(resourceType[(indexOfTypeSeparator + 1)..]);
+            }
```
JohnathonMohr, line 101: It appears resourceTypeSuffixes is not significantly used (only in two places):
1. Found resource matches prefix, and suffix is empty (lines 107+109) - this is equivalent to checking if the resource type matches the full `resourceType`.
2. prefix and suffix are concatenated together with '/' to search again recursively (line 116) - this just gets back the original `resourceType` string again.

You should be able to just build a list of 'parent' types and search into those resources, looking for the full original type string.

VeraBE: I'm not sure if I understand the suggestion (the last line), you just mean I shouldn't calculate the suffixes and maybe rename the prefixes to "parents"? If that's the case: I agree, the suffixes thing is a leftover from a previous solution, I didn't notice it doesn't make sense anymore

JohnathonMohr: Yes, that's what I meant.

docs/authoring-json-rules.md
```diff
@@ -43,7 +43,7 @@ Evaluation of ARM templates is performed on the JSON representation of the templ
 
 Since most rules apply only to specific types of Azure resources, the `resourceType` property gives rule authors a shorthand to only evaluate those types of resources.  If `resourceType` is specified, the path specified in `path` becomes relative to the resource selected in the template.
 
-The behavior of the `resourceType` property is to find a property called "resources" in the current scope that is an array of objects, look for a "type" property in each of the objects, and keep only the resources where the value of "type" matches the string in `resourceType`.  See [Scopes](#scopes) for more information on scopes.
+The behavior of the `resourceType` property is to find a property called "resources" in the current scope that is an array of objects, look for a "type" property in each of the objects, and keep only the resources where the full value of "type" matches the string in `resourceType`. Consider that if a resource of type `Microsoft.Sql/servers` has a child of type `auditingSettings`, then the full type of the child is actually `Microsoft.Sql/servers/auditingSettings`. See [Scopes](#scopes) for more information on scopes.
```
JohnathonMohr, line 46: Update to say that resources can also be matched within children of child resources at the current scope (you may need to significantly reword this statement into a few sentences).

src/Analyzer.Utilities/JsonPathResolver.cs
```diff
@@ -74,20 +75,44 @@ public IEnumerable<IJsonPathResolver> Resolve(string jsonPath)
...
+            
             foreach (var resource in resolvedTokens)
             {
-                if (string.Equals(resource.Value.InsensitiveToken("type")?.Value<string>(), resourceType, StringComparison.OrdinalIgnoreCase))
+                foreach (var resourceTypeParent in resourceTypeParents)
                 {
-                    yield return new JsonPathResolver(resource.Value, resource.Value.Path, this.resolvedPaths);
+                    if (string.Equals(resource.Value.InsensitiveToken("type")?.Value<string>(), resourceTypeParent, StringComparison.OrdinalIgnoreCase))
+                    {
+                        if (resourceTypeParent == resourceType)
```
JohnathonMohr, line 100: Instead of treating the actual resource being search for like a parent (including it with the list of parents), can we break this out so the code is a little clearer?

Psuedo-code:
```cs
var resourceTypeParents = new List<string>();
// Populate with only parent types

foreach (var resource in resolvedTokens)
{
    if (resource.Type == resourceType)
    {
        yield return new JsonPathResolver(resource);
    }
    else if (resourceTypeParents.Exists(r => r.Type == resource.Type))
    {
        // This resource is a type-parent of the resource being searched for.
        // Search children of this resource for the one in question.
        var parentScope = new JsonPathResolver(resource);
        foreach (var child in parentScope.ResolveResourceType(resourceType))
        {
            yield return child;
        }
    }
}
```

src/Analyzer.Utilities/JsonPathResolver.cs
```diff
@@ -74,20 +75,44 @@ public IEnumerable<IJsonPathResolver> Resolve(string jsonPath)
         /// <returns>An enumerable of resolvers with a scope of a resource of the specified type.</returns>
         public IEnumerable<IJsonPathResolver> ResolveResourceType(string resourceType)
```
JohnathonMohr, line 76: Add tests for new functionality.

src/Analyzer.Utilities/JsonPathResolver.cs
```diff
@@ -74,20 +75,41 @@ public IEnumerable<IJsonPathResolver> Resolve(string jsonPath)
...
+            }
+
             foreach (var resource in resolvedTokens)
             {
-                if (string.Equals(resource.Value.InsensitiveToken("type")?.Value<string>(), resourceType, StringComparison.OrdinalIgnoreCase))
+                if (resourceTypesAreEqual(resource, resourceType))
                 {
                     yield return new JsonPathResolver(resource.Value, resource.Value.Path, this.resolvedPaths);
                 }
+                else if (resourceTypeParents.Exists(parentResourceType => resourceTypesAreEqual(resource.Value, parentResourceType))) {
```
JohnathonMohr, line 105: nit: formatting
```suggestion
                else if (resourceTypeParents.Exists(parentResourceType => resourceTypesAreEqual(resource.Value, parentResourceType)))
                {
```

src/Analyzer.Utilities.UnitTests/JsonPathResolverTests.cs
```diff
@@ -14,6 +15,43 @@ namespace Microsoft.Azure.Templates.Analyzer.Utilities.UnitTests
...
+                                ""resources"": [
+                                    {
+                                        ""type"": ""Microsoft.ResourceProvider/resource2/resource3/resource4""
+                                    }    
+                                ]
+                            }
+                        ]
+                    }
+                ]
+            }", "Microsoft.ResourceProvider/resource2/resource3/resource4", new int[][] { new int[] { 1, 0, 0 } }, "1 Matching Child Resource" } }
```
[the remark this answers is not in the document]
JohnathonMohr: Oh, but Template Expansion won't actually be tested here...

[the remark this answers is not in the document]
JohnathonMohr: Hmm, I would have expected a failure in the TA-028 tests with a duplicate line number, but it apparently passes.  I wonder what I'm missing. 🤔

src/Analyzer.Core.BuiltInRuleTests/Tests/TA-000021/AutomationAccountsVariables.json
```diff
@@ -35,6 +35,22 @@
       "name": "automationAccounts/isEncryptedNotDefined",
```
JohnathonMohr, line 35: Add a resource of type "Microsoft.Automation/automationAccounts" that this resource depends on.  That will provide another test for resource selection.

src/Analyzer.Utilities/JsonPathResolver.cs
```diff
@@ -74,20 +75,42 @@ public IEnumerable<IJsonPathResolver> Resolve(string jsonPath)
...
             if (!resolvedPaths.TryGetValue(fullPath, out var resolvedTokens))
             {
                 var resources = this.currentScope.InsensitiveTokens("resources[*]");
                 resolvedTokens = resources.Select(r => (FieldContent)r).ToList();
                 resolvedPaths[fullPath] = resolvedTokens;
             }
 
+            // When trying to resolve Microsoft.Web/sites/siteextensions, for example, we should consider that siteextensions can be a child resource of Microsoft.Web/sites (a parent of the original resource type)
+            var resourceTypeParents = new List<string> { };
+            var indexesOfTypesSeparators = Regex.Matches(resourceType, "/").Cast<Match>().Select(m => m.Index).Skip(1);
```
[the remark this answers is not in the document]
nonik0: as MatchCollection implements `IEnumerable<Match>`

nonik0 (approved):

src/Analyzer.Core.BuiltInRuleTests/Tests/TA-000021/AutomationAccountsVariables.json
VeraBE: Created #261

PR #277, opened 2022-08-18
Run security rules from PSRule by default, and exclude repeated rules

VeraBE:
## Description

Added a baseline to run the PSRule rules under the [Security pillar](https://azure.github.io/PSRule.Rules.Azure/en/rules/module/#security) by default, except the ones already present in TemplateAnalyzer; and another one to exclude those repeated rules when using `--run-all-rules`

Documentation on PSRule baselines: https://microsoft.github.io/PSRule/v2/concepts/PSRule/en-US/about_PSRule_Baseline/

---


README.md
```diff
@@ -35,7 +35,7 @@ Argument | Description
 **(Optional)** `--report-format` | Valid formats:<br/>*Console*: output results to the console in plain text. **(default)**<br/>*Sarif*: output results to a file in [SARIF](https://sarifweb.azurewebsites.net) format.
 `-o` or `--output-file-path` | **(Required if `--report-format` is *Sarif*)**  File path to output SARIF results to.
 **(Optional)** `-v` or `--verbose` | Shows details about the analysis
-**(Optional)** `--run-powershell` | Also runs the PowerShell-based rules
+**(Optional)** `--run-all-rules` | Run all the rules against the templates, including non-security ones
```
JohnathonMohr, line 38: How about '--include-non-security-rules' instead?  On first glance it seems odd, and I wonder "wait, why do I have to manually specify I want to run them all?"  Having "non-security" specified up-front clarifies it.

JohnathonMohr: Also, this flag only applies to PSRule at the moment.  This might get complex if we introduce non-security checks in JSON rules as well.

docs/customizing-evaluation-outputs.md
```diff
@@ -1,7 +1,7 @@
 # Customizing Evaluation Outputs
 
 ## Overview
-Template BPA customization is authored in JSON.  Custom evaluation settings are written in a separate configuration and used on tool execution. See the main README for details on [how to run Template BPA with a configuration file](https://github.com/Azure/template-analyzer#using-the-template-bpa). 
+Template BPA customization is authored in JSON. Custom evaluation settings are written in a separate configuration and used on tool execution. See the main README for details on [how to run Template BPA with a configuration file](https://github.com/Azure/template-analyzer#using-the-template-bpa). It currently only supports the JSON-based rules.
```
JohnathonMohr, line 4:
```suggestion
Using an additional JSON configuration file, the Template BPA can be customized in how it runs JSON-based rules.  For example, specific rules can be included in/excluded from execution, or the severity of a rule can be changed. See the main README for details on [how to run Template BPA with a configuration file](https://github.com/Azure/template-analyzer#using-the-template-bpa).

**Note:** The configuration file only affects JSON-based rules; PSRule rules cannot be configured here at this time.
```

src/Analyzer.Core.UnitTests/TemplateAnalyzerTests.cs
```diff
@@ -115,26 +104,14 @@ public void AnalyzeTemplate_BicepTemplateIsInvalid_ThrowTemplateAnalyzerExceptio
...
             }
             finally
             {
                 File.Delete(templateFilePath);
             }
         }
 
-        [TestMethod]
-        [ExpectedException(typeof(TemplateAnalyzerException))]
-        public void AnalyzeTemplate_MissingFilePathWithPowerShellOn_ThrowTemplateAnalyzerException()
```
JohnathonMohr, line 128: Why was this test removed?

src/Analyzer.PowerShellRuleEngine.UnitTests/Analyzer.PowerShellRuleEngine.UnitTests.csproj
```diff
@@ -24,4 +24,8 @@
     </Content>
   </ItemGroup>
 
+  <ItemGroup>
+    <None Remove="templates\excluded_rules.json" />
```
JohnathonMohr, line 28: Was this meant to be deleted?  It's still added in the PR.

src/Analyzer.PowerShellRuleEngine.UnitTests/PowerShellRuleEngineTests.cs
```diff
@@ -69,32 +82,97 @@ public void AnalyzeTemplate_ValidTemplate_ReturnsExpectedEvaluations(string temp
...
+            else
+            {
+                var powerShellRuleEngineSecurityRules = new PowerShellRuleEngine(false);
+                evaluations = powerShellRuleEngineSecurityRules.AnalyzeTemplate(templateContext);
+            }
+
+            Assert.IsTrue(!evaluations.Any(evaluation => evaluation.RuleId == "AZR-000081")); // It's the id of Azure.AppService.RemoteDebug
+
+            // The RepeatedRulesBaseline will only be used when all rules are run:
+            if (runsAllRules)
```
JohnathonMohr, line 127: Should this be removed?  Otherwise line 143 will never be hit.

src/Analyzer.PowerShellRuleEngine.UnitTests/Analyzer.PowerShellRuleEngine.UnitTests.csproj
JohnathonMohr: Ah, it's used in a test.  Please remove this change.

src/Analyzer.PowerShellRuleEngine.UnitTests/PowerShellRuleEngineTests.cs
```diff
@@ -69,32 +82,97 @@ public void AnalyzeTemplate_ValidTemplate_ReturnsExpectedEvaluations(string temp
...
+                var matchingEvaluation = failedEvaluations.Find(evaluation => evaluation.Result.LineNumber == expectedLineNumber);
+                failedEvaluations.Remove(matchingEvaluation);
+            }
+            Assert.IsTrue(failedEvaluations.IsEmptyEnumerable());
+        }
+
+        [DataTestMethod]
+        [DataRow(true, DisplayName = "Repeated rules are excluded when running all the rules")]
+        [DataRow(true, DisplayName = "Repeated rules are excluded when running only the security rules")]
+        public void AnalyzeTemplate_ValidTemplate_ExcludesRepeatedRules(bool runsAllRules)
```
JohnathonMohr, line 97: I'm not sure what's supposed to be tested here.  The title describes repeated rules being excluded, but the only assertion in the test is whether or not a specific rule appears, which is related to whether or not non-security rules are run, unaffected by repetition.

src/Analyzer.PowerShellRuleEngine/PowerShellRuleEngine.cs
```diff
@@ -74,13 +82,37 @@ public IEnumerable<IEvaluation> AnalyzeTemplate(TemplateContext templateContext)
...
+                        Outcome = RuleOutcome.Fail,
+                        Culture = new string[] { "en-US" } // To avoid warning messages when running tests in Linux
+                    },
+                    Include = new IncludeOption
+                    {
+                        Path = new string[]
+                        {
+                            ".ps-rule",
+                            Path.Combine(Path.GetDirectoryName(AppContext.BaseDirectory), "baselines", "SecurityBaseline.Rule.json"),
+                            ".ps-rule",
```
JohnathonMohr, line 94: Duplicate?

src/Analyzer.Reports.UnitTests/SarifReportWriterE2ETests.cs
```diff
@@ -40,32 +40,49 @@ public void AnalyzeTemplateTests()
             }
 
             // assert
```
JohnathonMohr, line 42: After these changes, the E2E tests no longer validate the scenario where a single rule has multiple distinct failures.  Previously, TA-000028 had two failing results, each with 3 failing lines.

src/Analyzer.Core.UnitTests/TemplateAnalyzerTests.cs
VeraBE: It was there to validate a check that doesn't exist anymore, the file path now is a required parameter of the function, it will always fail if it's not specified

src/Analyzer.PowerShellRuleEngine.UnitTests/PowerShellRuleEngineTests.cs
VeraBE: There are two assertions in the code. One checks that whether running all the rules or not an excluded rule is not executed, and the second one validates that when the exclusion baseline is not defined properly the rule is indeed triggering with the test file

README.md
VeraBE: Thoughts from someone else before I refactor the name everywhere?

src/Analyzer.Core.UnitTests/TemplateAnalyzerTests.cs
JohnathonMohr: > it will always fail if it's not specified

I would argue that behavior should be validated with this test.

src/Analyzer.PowerShellRuleEngine.UnitTests/PowerShellRuleEngineTests.cs
JohnathonMohr: I see.  Can you add a comments explaining more of how the test works? It would be worth mentioning that the specific id you're looking for is a duplicate of a JSON rule, and should be excluded, unless the baseline file is missing (moved in the test).

src/Analyzer.Core.UnitTests/TemplateAnalyzerTests.cs
VeraBE: Do we validate for all our functions that they fail when we don't pass a required argument? :P

JohnathonMohr: We should.  If it's required, we should validate that the code enforces the requirement.  Example tests for LeafExpression: https://github.com/Azure/template-analyzer/blob/development/src/Analyzer.JsonRuleEngine.UnitTests/LeafExpressionTests.cs#L144#L156

JohnathonMohr: Put another way, if it's required, but there's a logical path in code where that requirement is somehow bypassed, there's a bug.  So the requirement should be tested.

VeraBE: I think we are not understanding each other, that test doesn't even compile anymore, because the function call is missing a parameter

VeraBE: I think we are not understanding each other, that test doesn't even compile anymore, because the function call is missing a parameter. It doesn't make sense to test what happens when the parameter is missing, because it doesn't compile when that happens

JohnathonMohr: Previously it was set to null by default, so it could be "left out".  The test should pass `null` as the value and expect a failure.

JohnathonMohr: Maybe "missing" vs. "null" is the confusing piece.  It should validate a failure from passing null as the file path.

README.md
BernieWhite: @JohnathonMohr @VeraBE Alternative suggestion:

`--baseline` or `--with-baseline`

PSRule has many baselines. By default, customers get the latest security baseline for BPA or they can specify one. 

For example:

`--baseline Azure.All` or `--baseline Azure.Preview`

BernieWhite (approved):
@VeraBE Just one additional comments otherwise all good.

src/Analyzer.PowerShellRuleEngine.UnitTests/PowerShellRuleEngineTests.cs
```diff
@@ -42,7 +47,15 @@ public void AnalyzeTemplate_ValidTemplate_ReturnsExpectedEvaluations(string temp
...
             };
 
-            var evaluations = powerShellRuleEngine.AnalyzeTemplate(templateContext);
+            IEnumerable<IEvaluation> evaluations;
+            if (runsAllRules)
+            {
+                evaluations = powerShellRuleEngineAllRules.AnalyzeTemplate(templateContext);
+            } else
+            {
+                var powerShellRuleEngineSecurityRules = new PowerShellRuleEngine(false);
```
anaismiller, line 56: Why don't you put this at the top too?

src/Analyzer.PowerShellRuleEngine.UnitTests/PowerShellRuleEngineTests.cs
```diff
@@ -15,19 +17,22 @@ namespace Microsoft.Azure.Templates.Analyzer.RuleEngines.PowerShellEngine.UnitTe
...
         [AssemblyInitialize]
         public static void AssemblyInitialize(TestContext context)
         {
-            powerShellRuleEngine = new PowerShellRuleEngine();
+            powerShellRuleEngineAllRules = new PowerShellRuleEngine(true);
         }
 
         [DataTestMethod]
-        [DataRow("template_and_resource_level_results.json", 12, new int[] { 1, 1, 1, 1, 8, 14, 17, 1, 17, 17, 1, 17 }, DisplayName = "Template with errors reported in both analysis stages")]
+        [DataRow("template_and_resource_level_results.json", true, 12, new int[] { 1, 1, 1, 1, 8, 14, 17, 1, 17, 17, 1, 17 }, DisplayName = "Running all the rules against a template with errors reported in both analysis stages")]
```
anaismiller, line 29: What does `in both analysis stages` mean?

anaismiller: oh tool errors and rule violations?

src/Analyzer.PowerShellRuleEngine.UnitTests/PowerShellRuleEngineTests.cs
```diff
@@ -69,32 +82,89 @@ public void AnalyzeTemplate_ValidTemplate_ReturnsExpectedEvaluations(string temp
...
+            }
+            Assert.IsTrue(failedEvaluations.IsEmptyEnumerable());
+        }
+
+        [DataTestMethod]
+        [DataRow(true, DisplayName = "Repeated rules are excluded when running all the rules")]
+        [DataRow(true, DisplayName = "Repeated rules are excluded when running only the security rules")]
+        public void AnalyzeTemplate_ValidTemplate_ExcludesRepeatedRules(bool runsAllRules)
+        {
+            var templateFilePath = Path.Combine(templatesFolder, "excluded_rules.json");
```
anaismiller, line 99: Can you rename the file to TestExcludedRules.json?

anaismiller: I thought it was a rule file or configuration file and it confused me.

src/Analyzer.Cli.FunctionalTests/CommandLineParserTests.cs
```diff
@@ -243,10 +243,18 @@ public void FilterRules_ValidConfig_RulesFiltered(bool isBicep, string configNam
...
                 if (specifyInCommand)
+                {
                     args = args.Concat(new[] { "--config-file-path", configName }).ToArray();
+                }
+
+                using StringWriter outputWriter = new();
+                Console.SetOut(outputWriter);
 
                 result = _commandLineParser.InvokeCommandLineAPIAsync(args);
-                Assert.AreEqual((int)ExitCode.Success, result.Result);
```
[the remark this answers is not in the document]
VeraBE: It doesn't work anymore because all those files return PSRule violations :/

src/Analyzer.PowerShellRuleEngine.UnitTests/PowerShellRuleEngineTests.cs
VeraBE: No, PSRule detects errors when looking at the whole file (through the file path), and when looking at each resource (`pipeline.Process(resource);`)

src/Analyzer.PowerShellRuleEngine.UnitTests/PowerShellRuleEngineTests.cs
VeraBE: Do you think it's necessary even considering it is in the "templates" folder?

src/Analyzer.Cli.FunctionalTests/CommandLineParserTests.cs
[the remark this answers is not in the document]
anaismiller: It does not have to be ExitCode.Success. And the value can be passed in as a parameter in the test function if each one is different

src/Analyzer.PowerShellRuleEngine.UnitTests/PowerShellRuleEngineTests.cs
anaismiller: Oh! Maybe this is something that should be added to the ReadMe? Not sure.

src/Analyzer.PowerShellRuleEngine.UnitTests/PowerShellRuleEngineTests.cs
anaismiller: If I plan to make a change to this test, it could help me save time understanding what it going on. And if I want to use this file in a test, would also help. 
If there is a different name, could also be good.

README.md
VeraBE: I'm not sure if I like the idea of having a config file to filter the JSON rules and the baseline file for the PSRule rules, shouldn't we try to keep a consistent interface for all types of rules in the tool? Maybe by translating the config file into baseline files? Even though it sounds a bit annoying because in our config file we use rule ids and the baseline needs rule names

JohnathonMohr: I agree, we *should* keep a consistent interface for configuring all rules, regardless of where/how they're written.  That's going to require a thoughtful design though, and shouldn't block our current progress.  For now, I think ignoring the available baselines and just keeping what we have (running just security rules by default, but allowing a way to additionally run non-security rules if desired) is best; then we can figure out how to continue the integration next.

anaismiller (approved):

src/Analyzer.Core/TemplateAnalyzer.cs
```diff
@@ -47,10 +47,10 @@ private TemplateAnalyzer(JsonRuleEngine jsonRuleEngine, PowerShellRuleEngine pow
         /// <summary>
         /// Creates a new <see cref="TemplateAnalyzer"/> instance with the default built-in rules.
         /// </summary>
-        /// <param name="usePowerShell">Whether or not to use PowerShell rules to analyze the template.</param>
+        /// <param name="runAllRules">Whether or not to run all the rules against the template.</param>
         /// <param name="logger">A logger to report errors and debug information</param>
         /// <returns>A new <see cref="TemplateAnalyzer"/> instance.</returns>
-        public static TemplateAnalyzer Create(bool usePowerShell, ILogger logger = null)
+        public static TemplateAnalyzer Create(bool runAllRules, ILogger logger = null)
```
JohnathonMohr, line 53: Rename here as well.

src/Analyzer.PowerShellRuleEngine/PowerShellRuleEngine.cs
```diff
@@ -27,9 +33,11 @@ public class PowerShellRuleEngine : IRuleEngine
         /// <summary>
         /// Creates a new instance of a PowerShellRuleEngine.
         /// </summary>
+        /// <param name="runAllRules">Whether or not to run all the rules against the template.</param>
         /// <param name="logger">A logger to report errors and debug information.</param>
-        public PowerShellRuleEngine(ILogger logger = null)
+        public PowerShellRuleEngine(bool runAllRules, ILogger logger = null)
```
JohnathonMohr, line 38: Rename here and class variable too.

src/Analyzer.PowerShellRuleEngine/Analyzer.PowerShellRuleEngine.csproj
```diff
@@ -8,13 +8,19 @@
   
   <ItemGroup>
     <PackageReference Include="Microsoft.PowerShell.SDK" Version="7.2.4" />
-    <PackageReference Include="Microsoft.PSRule.Rules.Azure" Version="1.18.0-B0010" />
-    <PackageReference Include="Microsoft.PSRule.SDK" Version="2.3.0-B0130" />
+    <PackageReference Include="Microsoft.PSRule.Rules.Azure" Version="1.19.0-B0010" />
```
JohnathonMohr, line 11: Please update the nuspec file for Analyzer.Core too.  We eventually need a better way to maintain that (created tracking issue #279), but for now, we need to update manually.

JohnathonMohr (changes requested):

src/Analyzer.PowerShellRuleEngine/Analyzer.PowerShellRuleEngine.csproj
JohnathonMohr: There are also other out-of-date versions in that nuspec, as well as dependencies that are missing entirely (Bicep).  Can you update those as well?

JohnathonMohr (approved):

PR #290, opened 2022-09-15
Bicep modules: Support for scanning and outputting results

### LINKED ISSUE #228: Deduplication of reported lines in analysis results

Sometimes the same line in a template is reported for a failure of the same rule.  This can happen for many valid reasons:
- Expected properties are absent, and the line number of the closest parent object declaration is the same for the missing properties
- A property can meet one of many conditions, but it doesn't meet any of them, resulting in a failure of all conditions
- A resource is copied, each copy fails a rule, and the line number in the original template would be the same for all copies

In many cases these failures would be distinguished from each other by better messages in the results.  However, while that ability is missing now, they appear to simply be duplicates and result in noise in the reports.

This can be solved broadly by implementing [IEquatable<T>](https://docs.microsoft.com/en-us/dotnet/api/system.iequatable-1?view=net-5.0) in classes implementing `IResult`, implementing both `Equals` and `GetHashCode` and comparing the line numbers across `IResult`s.  Reports could then do something simple like the following:
``` cs
IEnumerable<IResult> distinctFailures = GetFailedResults(evaluation).Distinct();
```
... which would filter out the apparent duplicates.  In the future when more detailed messaging is implemented to show information specific to a particular line, this solution would continue to "just work" in the reports by simply updating the `IResult` classes to compare the messages when determining equality.  For example, this would result in the same line being shown in multiple *location* properties of SARIF, but with different messages, so it's clear there is more than one failure at that line.

(A little more effort would be needed to address duplicate lines across distinct results in SARIF, for example in the resource copy scenario, since they would be in separate `Evaluation`s, but this scenario is still greatly simplified by the suggested change.)

### Additional context

_No response_

### LINKED ISSUE #275: Repeated code "GetFailedLines" function in several parts of the project

There is a repeated function "GetAllFailedLines" in TestRunner.cs, "GetFailedLines" in ConsoleReportWriter.cs, and "GetFailedLines" in TemplateAnalyzerTests.cs. 
They implement the exact same logic. 

### Additional context

_No response_

nonik0:
## Description
This PR adds support for bicep modules, using the recently added support for nested templates. Changes include:
- Updates to source location resolution to handle results not in target file
- Updates and refactoring of reporting functionality for writing results from modules
- PSRuleEngine updated to use BicepSourceLocationResolver when applicable
- Various tests for updates and new functionality
 
Fixes #275

---


src/Analyzer.JsonRuleEngine/JsonRuleResult.cs
```diff
@@ -30,5 +31,22 @@ internal class JsonRuleResult : IResult
         /// Gets the expression associated with this result
         /// </summary>
         internal Expression Expression { get; set; }
+
+        public bool Equals(JsonRuleResult other)
```
nonik0, line 35: All results have this identical code--create an abstract Result class so this functionality can just be shared and not duplicated in each implementation?

src/Analyzer.Reports.UnitTests/TestCases.cs
```diff
@@ -70,7 +72,7 @@ public class TestCases
                         HelpUri = "https://domain.com/help",
                         Passed = false,
                         Evaluations = Enumerable.Empty<MockEvaluation>(),
-                        Result = new MockResult { Passed = false, LineNumber = 65 }
```
anaismiller, line 73: Is it useful to keep the words `LineNumber = `?

src/Analyzer.Types/SourceLocation.cs
```diff
@@ -0,0 +1,71 @@
...
+        public override bool Equals(object obj)
+        {
+            var result = obj as SourceLocation;
+            return (result != null) && Equals(result);
+        }
+
+        /// <summary>
+        /// Compare with source location for equality
+        /// </summary>
+        /// <param name="other"></param>
```
anaismiller, line 63: Do you want to rename `other` to `sourceLocation`?

src/Analyzer.Utilities/ISourceLocationResolver.cs
```diff
@@ -1,20 +1,20 @@
...
     {
-        // TODO: include source file for Bicep and rename to ISourceLocationResolver once nested templates are supported
-
         /// <summary>
         /// Find the line number of a path.
         /// </summary>
         /// <param name="path">The path describing the location in the document.</param>
-        /// <returns>The line number of the specified path.</returns>
-        public int ResolveLineNumber(string path);
+        /// <returns>TODO</returns>
```
anaismiller, line 17: todo

src/Analyzer.Core/TemplateAnalyzer.cs
```diff
@@ -117,7 +117,7 @@ public IEnumerable<IEvaluation> AnalyzeTemplate(string template, string template
             var  evaluations = AnalyzeAllIncludedTemplates(template, parameters, templateFilePath, templateContext, "");
 
             // For each rule we don't want to report the same line more than once
```
VeraBE, line 119: This code is a temporary solution until issue #228 is resolved, it shouldn't be needed if the issue is closed by this PR. It also shouldn't be needed to dedupe results in ConsoleReportWriterTests, SarifReportWriterTests and TemplateAnalyzerTests

src/Analyzer.Reports.UnitTests/ConsoleReportWriterTests.cs
```diff
@@ -17,9 +19,9 @@ public class ConsoleReportWriterTests
     {
         [DataTestMethod]
         [DynamicData("UnitTestCases", typeof(TestCases), DynamicDataSourceType.Property, DynamicDataDisplayName = "GetTestCaseName", DynamicDataDisplayNameDeclaringType = typeof(TestCases))]
-        public void WriteResults_Evalutions_ReturnExpectedSarifLog(string _, MockEvaluation[] evaluations)
+        public void WriteResults_Evalutions_ReturnExpectedConsoleLog(string _, MockEvaluation[] evaluations)
```
VeraBE, line 22:
```suggestion
        public void WriteResults_Evaluations_ReturnExpectedConsoleLog(string _, MockEvaluation[] evaluations)
```

src/Analyzer.Core.BuiltInRuleTests/TestRunner.cs
```diff
@@ -74,7 +74,7 @@ private IEnumerable<int> GetAllFailedLines(IEvaluation evaluation)
         {
             if (!evaluation.Result?.Passed ?? false)
             {
-                yield return evaluation.Result.LineNumber;
+                yield return evaluation.Result.SourceLocation.LineNumber;
```
VeraBE, line 77: Shouldn't this whole function be removed and its usage replaced by a statement using the new ReportsHelper.GetFailedResults function?

src/Analyzer.Reports.UnitTests/SarifReportWriterE2ETests.cs
```diff
@@ -116,39 +127,68 @@ public void AnalyzeTemplateTests()
...
-                ? UriHelper.MakeValidUri(Path.Combine(secondTemplatePathPieces.Append(secondTemplateName).ToArray()))
+                ? UriHelper.MakeValidUri(Path.Combine(secondTemplatePathPieces.Append(secondTemplate).ToArray()))
                 : new Uri(secondTemplatePath).AbsoluteUri;
 
             // act
             var memStream = new MemoryStream();
-            using (var writer = SetupWriter(memStream, targetDirectory))
+            try
             {
+                if (Directory.Exists(targetDirectory))
```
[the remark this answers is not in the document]
nonik0: The second directory will always be equal or under the first directory so it's just an optimization. I can add a comment to clarify.

src/Analyzer.Reports.UnitTests/SarifReportWriterE2ETests.cs
```diff
@@ -161,36 +201,37 @@ public void AnalyzeDirectoryTests(bool secondTemplateUsesRelativePath, params st
...
                     })
                 },
                 { "TA-000028", (
-                    file: expectedSecondTemplateFilePathInSarif,
+                    file: secondIsBicep ? nestedTemplate : expectedSecondTemplateFilePathInSarif,
                     uriBase: secondTemplateUsesRelativePath ? SarifReportWriter.UriBaseIdString : null,
                     lines: new List<List<int>> {
-                        new List<int> { 23, 24, 25 },
-                        new List<int> { 43, 44, 45 }
+                        secondIsBicep ? new List<int> { 14, 15, 16 } : new List<int> { 23, 24, 25 },
```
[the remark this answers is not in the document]
nonik0: Yes I figure once we get to that point we'll likely overhaul the tests..

src/Analyzer.Reports.UnitTests/SarifReportWriterE2ETests.cs
```diff
@@ -161,36 +201,37 @@ public void AnalyzeDirectoryTests(bool secondTemplateUsesRelativePath, params st
             run.OriginalUriBaseIds.Count.Should().Be(1);
             run.OriginalUriBaseIds["ROOTPATH"].Uri.Should().Be(new Uri(targetDirectory, UriKind.Absolute));
 
+            var firstIsBicep = (firstTemplateFileInfo.Extension == ".bicep");
+            var secondIsBicep = (secondTemplateFileInfo.Extension == ".bicep");
```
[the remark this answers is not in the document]
nonik0: Ah this is vestigial code...yes for now the templates will be the same, but once I fix the ARM JSON as a bicep module line number bug, I could remove the bicep version of SQLServerAuditingSettings.json...but still in that case the second template would always be JSON. Will simplify.

src/Analyzer.Reports.UnitTests/TestCases.cs
nonik0: Could you clarify? This is just updating all results from line number to source location in test cases.

src/Analyzer.Types/SourceLocation.cs
```diff
@@ -0,0 +1,71 @@
...
+        /// </summary>
+        /// <param name="filePath">File path of the source location</param>
+        /// <param name="lineNumber">Line number of source location</param>
+        /// <param name="referencedBy">Source location that refers to this location (i.e. line for bicep module)</param>
+        public SourceLocation(string filePath, int lineNumber, SourceLocation referencedBy = null)
+        {
+            this.FilePath = filePath;
+            this.LineNumber = lineNumber;
+            this.ReferencedBy = referencedBy;
+            //this.ActualLocation = this.GetActualLocation();
```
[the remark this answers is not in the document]
nonik0: yup

src/Analyzer.Core/TemplateAnalyzer.cs
nonik0: OK that makes sense I will try removing. I still think that the deduping of the results in the tests is needed because a result with multiple paths (i.e. bicep modules) will have multiple evaluations with identical results (except they technically differ by target file/reference path).

This is easy to verify for sure though by removing it and seeing if the tests still pass.

src/Analyzer.Core.BuiltInRuleTests/TestRunner.cs
nonik0: yes it should

src/Analyzer.JsonRuleEngine/JsonRuleResult.cs
nonik0: This might be a little trickier than I thought since if I do this the simple way with the abstract class impl, a powershell and json rule engine result on the same line could be duped if I don't preserve the type comparison of each derived type (i.e I need to make sure the type of the of result factors in the equality as well when comparing). I am not sure yet if that matters though.

src/Analyzer.Core/TemplateAnalyzer.cs
VeraBE: I'm not sure if I follow, why identical results? Like the same line number but a different file path?

src/Analyzer.Reports.UnitTests/SarifReportWriterE2ETests.cs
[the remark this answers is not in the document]
anaismiller: Oh, ok

src/Analyzer.Reports.UnitTests/TestCases.cs
anaismiller: I think it was updated or I mis-saw. I don't notice it anymore.

src/Analyzer.Reports/ReportsHelper.cs
```diff
@@ -0,0 +1,121 @@
...
+            passedEvaluations = 0;
+
+            foreach (var evaluation in evaluations)
+            {
+                if (!evaluation.Passed)
+                {
+                    (var actualFile, var failedResults) = GetResultsByFileInternal(evaluation);
+
+                    // a file's results may have already been output if analyzing directory
+                    if (filesToSkip.Contains(actualFile))
```
JohnathonMohr, line 50: Instead of skipping evaluations that are from the same file, let's just not analyze files that have already been analyzed.  This will make execution way more performant in directories that have lots of Bicep modules in use.  The only scenario we need to be cautious of is if parameters are provided when analyzing, so we should reevaluate the same file if different parameters are passed.

It should be as simple as
1. Track file identifiers analyzed in `TemplateAnalyzer.AnalyzeTemplate` (with the parameters used in analysis), and return the evaluations already found if it's been scanned before.
2. Track the same thing in the CLI, but simply don't call `TemplateAnalyzer.AnalyzeTemplate` for a file that's already been analyzed.

(2) should be done to optimize the most common use case (executing from the command line).  (1) should be done just to benefit other scenarios where Analyzer.Core is used directly in another project outside BPA.  This isn't as high of a priority.

src/Analyzer.Reports/ReportsHelper.cs
```diff
@@ -0,0 +1,121 @@
...
+        /// <summary>
+        /// Gets all failed results in an evaluation, sorts, and returns rule ID
+        /// </summary>
+        /// <param name="evaluation"></param>
+        /// <returns></returns>
+        /// <exception cref="Exception"></exception>
+        public static (string, IList<IResult>) GetResultsByFileInternal(Types.IEvaluation evaluation)
+        {
+            // get all distinct failed results in evaluation
+            var failedResults = GetFailedResultsAsList(evaluation).Distinct().ToList();
```
JohnathonMohr, line 90: This should probably be implemented in the Evaluation: `evaluation.FailedResults()`, which can return distinct failed results.  Using a `HashSet<Result>` will also get rid of the need to call `Distinct()`.

JohnathonMohr (commented):
Initial review

src/Analyzer.JsonRuleEngine/JsonRuleResult.cs
JohnathonMohr: A couple thoughts:
- If type does matter, you can likely just include that as another check in the implementation.  The concrete type should come through at runtime, even if implemented in the abstract base.
- It's also possible we don't need to keep separate implementations for JSON and PowerShell.  Maybe we can merge them.

src/Analyzer.Cli.FunctionalTests/Tests/TemplateWithModule.bicep
```diff
@@ -3,4 +3,11 @@ module testModule 'AppServicesLogs-Failures.bicep' = {
 	params: {
 		location: 'testLocation'
 	}
-}
\ No newline at end of file
+}
+
+module hw 'br/public:samples/hello-world:1.0.2' = {
```
JohnathonMohr, line 8: What is this reference?  Is this actually downloaded?  How does this impact the template analysis?

src/Analyzer.Reports/ConsoleReportWriter.cs
```diff
@@ -5,7 +5,10 @@
 using System.Collections.Generic;
 using System.IO.Abstractions;
 using System.Linq;
+using System.Reflection;
+using System.Runtime.CompilerServices;
 using Microsoft.Azure.Templates.Analyzer.Types;
+using Microsoft.CodeAnalysis.Sarif;
```
JohnathonMohr, line 11: I think these usings are unneeded?

src/Analyzer.Reports/SarifReportWriter.cs
```diff
@@ -7,8 +7,10 @@
 using System.IO;
 using System.IO.Abstractions;
 using System.Linq;
+using System.Runtime.CompilerServices;
 using Microsoft.Azure.Templates.Analyzer.Types;
 using Microsoft.CodeAnalysis.Sarif;
+using Microsoft.CodeAnalysis.Sarif.Baseline;
```
JohnathonMohr, line 13: Remove unnecessary usings.

src/Analyzer.Reports.UnitTests/SarifReportWriterE2ETests.cs
```diff
@@ -93,10 +104,10 @@ public void AnalyzeTemplateTests()
                 result.Locations.Count.Should().Be(expectedLines.Count);
                 foreach (var location in result.Locations)
                 {
-                    location.PhysicalLocation.ArtifactLocation.Uri.OriginalString.Should().BeEquivalentTo(artifactUriString);
-                    var line = location.PhysicalLocation.Region.StartLine;
+                    location.PhysicalLocation.ArtifactLocation.Uri.OriginalString.Should().BeEquivalentTo(evalFile);
```
JohnathonMohr, line 107: This is simply verifying that the same file path is used for all locations, because `evalFile` originally comes from the first location - essentially it's asserting the location is equal to itself for the first iteration in this loop.  Is that the intent?  And is this strictly necessary?  If you think it is, add a comment clarifying that, because it looks strange.

src/Analyzer.Reports.UnitTests/SarifReportWriterTests.cs
```diff
@@ -138,25 +126,39 @@ private void AssertSarifLog(SarifLog sarifLog, IEnumerable<Types.IEvaluation> ev
...
                     result.RuleId.Should().BeEquivalentTo(evaluation.RuleId);
                     result.Message.Id.Should().BeEquivalentTo("default");
                     result.Level.Should().Be(FailureLevel.Error);
 
-                    var evalResults = new List<Types.IResult>();
-                    TraverseResults(evalResults, evaluation);
-                    result.Locations.Count.Should().Be(evalResults.Count);
+                    if (evalResults.First().SourceLocation.FilePath != templateFilePath.FullName)
+                    {
+                        result.AnalysisTarget.Uri.OriginalString.Should().BeEquivalentTo(templateFilePath.Name);
```
JohnathonMohr, line 148: Add validation that AnalysisTarget is null otherwise.

src/Analyzer.Reports.UnitTests/SarifReportWriterTests.cs
```diff
@@ -138,25 +126,39 @@ private void AssertSarifLog(SarifLog sarifLog, IEnumerable<Types.IEvaluation> ev
...
+                    }
+
                     for (int j = 0; j < evalResults.Count; j++)
                     {
-                        result.Locations[j].PhysicalLocation.ArtifactLocation.Uri.OriginalString.Should().BeEquivalentTo(templateFilePath.Name);
-                        result.Locations[j].PhysicalLocation.Region.StartLine.Should().Be(evalResults[j].LineNumber);
+                        var evalResultFileName = (new FileInfo(evalResults[j].SourceLocation.FilePath)).Name;
```
JohnathonMohr, line 153: A little cleaner in my opinion:
```suggestion
                        var evalResultFileName = Path.GetFileName(evalResults[j].SourceLocation.FilePath);
```

src/Analyzer.Reports.UnitTests/SarifReportWriterE2ETests.cs
```diff
@@ -84,6 +88,13 @@ public void AnalyzeTemplateTests()
                 expectedLinesForRun.ContainsKey(result.RuleId).Should().BeTrue("Unexpected result found in SARIF");
                 result.Level.Should().Be(FailureLevel.Error);
 
+                var evalFile = result.Locations.First().PhysicalLocation.ArtifactLocation.Uri.OriginalString;
+                var resultInOtherFile = evalFile != artifactUriString;
+                if (resultInOtherFile)
+                {
+                    result.AnalysisTarget.Uri.OriginalString.Should().BeEquivalentTo(artifactUriString);
```
JohnathonMohr, line 95: Add validation that AnalysisTarget is null otherwise.

src/Analyzer.Reports.UnitTests/SarifReportWriterE2ETests.cs
```diff
@@ -239,15 +280,21 @@ public void AnalyzeDirectoryTests(bool secondTemplateUsesRelativePath, params st
                 expectedLines.Should().NotBeNull("There shouldn't be a line number reported outside of the expected lines.");
                 linesForResult.Remove(expectedLines);
 
+                // Validate analysis target if result is from nested template
+                if (fileName == nestedTemplate)
+                {
+                    result.AnalysisTarget.Uri.OriginalString.Should().BeEquivalentTo(secondTemplate);
```
JohnathonMohr, line 286: Add validation that AnalysisTarget is null otherwise.

src/Analyzer.JsonRuleEngine/JsonRuleResult.cs
nonik0: I don't think it matters since all evaluations will have the same types of results, and when results are compared across expressions (to dedupe), the rule type of the expression will also be a part of that comparison (assuming rule IDs are unique to each result type). Regardless, I will implement the type checking of the concrete type.

src/Analyzer.BicepProcessor/BicepLocationResolver.cs
```diff
@@ -34,20 +41,47 @@ public BicepLocationResolver(TemplateContext templateContext)
...
+            // sort smallest to largest map size to sort into reference order
+            matches.Sort((x, y) => y.mapSize.CompareTo(x.mapSize));
+
+            // default to result from JSON if no matches (i.e. a bicep module references a JSON template that wouldn't be in source map)
+            if (matches.Count == 0)
+            {
+                return new SourceLocation(this.EntrypointFilePath, jsonLine + 1); // convert line number back to 1-indexing
+            }
+
+            // TODO: verify entrypoint file should always be top of call stack
```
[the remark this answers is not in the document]
nonik0: This was more of a assumption I was verifying during testing.

src/Analyzer.Core/TemplateAnalyzer.cs
nonik0: Yes, if a bicep file has multiple references to another template with results, there would be multiple "paths" through the bicep to those issues--so in the results we would have an expression with identical results for each unique path to that template.

src/Analyzer.Reports.UnitTests/SarifReportWriterE2ETests.cs
nonik0: This is simply checking what file was evaluated for this SARIF result, as each result will be for a single template. Will clarify in comments.

src/Analyzer.Reports/ReportsHelper.cs
nonik0: I initially skipped this optimization because I wasn't sure moving forward if we'd really want it or not--though if our goal is not to increase the granularity of data in our results (i.e. preserving more of the module information) then I suppose we could go ahead with this.

src/Analyzer.Reports/ReportsHelper.cs
```diff
@@ -0,0 +1,121 @@
...
+
+            return resultsByFile;
+        }
+
+        /// <summary>
+        /// Gets all failed results in an evaluation, sorts, and returns rule ID
+        /// </summary>
+        /// <param name="evaluation"></param>
+        /// <returns></returns>
+        /// <exception cref="Exception"></exception>
```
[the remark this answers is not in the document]
nonik0: vestigial testing, removing :)

src/Analyzer.BicepProcessor/BicepLocationResolver.cs
```diff
@@ -34,20 +41,47 @@ public BicepLocationResolver(TemplateContext templateContext)
...
+            if (Path.GetFileName(this.EntrypointFilePath) != matches.First().filePathRelativeToEntrypoint) throw new Exception();
```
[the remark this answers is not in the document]
nonik0: Yes, so overall I am not happy with this code as it's sort of hard to read. It requires a good deal of knowledge of the how the [SourceMap's ](https://github.com/Azure/bicep/blob/main/src/Bicep.Core/Emit/SourceAwareJsonTextWriter.cs) metadata is built. I potentially have a nicer solution using some bicep metadata but currently asking bicep team for some details about some gaps in that data.

If I have a result in some template that is referenced as a module, then each SourceMapFileEntry will have one SourceMapFileEntry where TargetLine is equal to the result line, if the template for that SourceMapFileEntry references the result. In the SourceMapFileEntry for the template that contains the result there will be a single mapping from the source to the target line, i.e. the source line ONLY maps to the one target line. But, in all of the files who indirectly reference the result through a module, there will be many SourceMapEntrys that match the same source line as the entry that maps to the target line, i.e. a source line with a module references ALL of the lines in the referenced template, including the line with the result.

So "mapCount" is how many SourceMapEntrys match the SourceLine of the SourceMapEntry that matches the TargetLine of the result. Sorry I am not sure how to say that more simply! By sorting by mapCount, you can get the "stack" of template references, where the highest mapCount will be the target file, and the lowest will be the file with the result.

I am happy to discuss further if needed.

src/Analyzer.Reports.UnitTests/SarifReportWriterE2ETests.cs
JohnathonMohr: Maybe this is a better request - can `evalFile` be determined without using the first location?  Otherwise, if all locations are wrong, the test might still pass because the "source of truth" is incorrect to start with.

src/Analyzer.Reports/ReportsHelper.cs
```diff
@@ -0,0 +1,121 @@
...
+
+                    // a file's results may have already been output if analyzing directory
+                    if (filesToSkip.Contains(actualFile))
+                    {
+                        continue;
+                    }
+
+                    if (!resultsByFile.ContainsKey(actualFile))
+                    {
+                        resultsByFile[actualFile] = new List<(IEvaluation, IList<IResult>)>();
```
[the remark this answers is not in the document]
nonik0: In order to do this I would probably implement a generic expression type to contain the IEnumerable<IExpression> that contains the unique results (and refactor the other methods to keep the outer expression of the IResults). Other alternatives creating instances of the existing evaluation types would be more involved I think.

So this would add some additional complexity/code I don't think really improves the readability/simplicity of the code overall since if requires a good amount of additional code/refactoring. Lmk if you want to discuss further.

src/Analyzer.Reports/ReportsHelper.cs
```diff
@@ -0,0 +1,121 @@
...
+                }
+                else
+                {
+                    passedEvaluations++;
+                }
+            }
+
+            foreach (var file in resultsByFile.Keys)
+            {
+                resultsByFile[file] = resultsByFile[file].Distinct().ToList();
```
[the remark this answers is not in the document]
nonik0: Yes you are right, good catch.

src/Analyzer.Reports.UnitTests/SarifReportWriterE2ETests.cs
nonik0: I see your point now--the location check in the foreach loop effectively does nothing since the evalFile value is taken from it originally. I noticed this I had an unused isReferenced argument for the test so I think I had this in mind at some point to fix the validation for the module reference case? 😅 Good catch, should be fixed with next commit..

src/Analyzer.Reports/ReportsHelper.cs
nonik0: Since IEvaluation is an interface, I did this using an extension method.

Also, the tests depend on GetFailedResults for getting the unduped line numbers (the spiritual successor to GetFailedLines), so I need to do the deduping outside of that method.

I got tripped up on this for a while because I thought I was getting different results from List.Distinct() and a HashSet when it was just the test dependency I forgot about. Would have been caught right away if I did the extension method first. 😅

src/Analyzer.Reports/ReportsHelper.cs
nonik0: OK this currently is actually not feasible because we do not have the "template reference tree" metadata to be able to determine the parent/child of any given analyzed template. This prevents us from filtering all the evaluations to the ones specific to each referenced module when tracking this data (i.e. all the evals that it directly or indirectly references through its own modules).

I think I could technically rebuild this relational data from the source map but it wouldn't be pretty or ideal. This is also a related gap affecting the ARM-template-as-bicep-module issue as well as the other comment on this PR regarding the BicepTemplateProcessor. I propose to create a bug for this work and I can do the work once we have the data available to us.

src/Analyzer.Cli.FunctionalTests/Tests/TemplateWithModule.bicep
nonik0: This is the [bicep public module registry](https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/modules#public-module-registry), so it actually downloaded a bicep file and includes a resource in the resultant ARM template.

This was originally in the tests (I think this file was renamed?) to verify bicep files with modules did not have any issues, even though we did not process them. Now this is actually downloaded we should at least have a test to verify this works with no issues. I don't think validation beyond that is necessary. Let me know your thoughts here, I've added comments to clarify what this does in the template.

src/Analyzer.JsonRuleEngine/JsonRuleResult.cs
nonik0: I ended up merging all the results into one class. Let me know your thoughts, there's some JSON engine specific stuff that makes it a bit less clean.

src/Analyzer.BicepProcessor/BicepLocationResolver.cs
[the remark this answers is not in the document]
VeraBE: Was it verified? Should the comment be removed?

src/Analyzer.Core.BuiltInRuleTests/Tests/TA-000001/AppServicesLogs-Failures.json
```diff
@@ -96,7 +96,7 @@
         {
           "apiVersion": "2019-08-01",
           "type": "config",
-          "name": "configAsChildResource",
+          "name": "web",
```
VeraBE, line 99: Was this renamed by mistake?

src/Analyzer.Core.BuiltInRuleTests/Tests/TA-000001/AppServicesLogs-Passes.json
```diff
@@ -122,7 +122,7 @@
         {
           "apiVersion": "2019-08-01",
           "type": "config",
-          "name": "configAsChildResource",
+          "name": "web",
```
VeraBE, line 125: Also renamed by mistake?

src/Analyzer.Core.UnitTests/TemplateAnalyzerTests.cs
```diff
@@ -41,9 +42,91 @@ public void AnalyzeTemplate_ValidInputValues_ReturnCorrectEvaluations(string res
...
+                        if (!failedSourceLocations.Any(existingLocation => existingLocation.SequenceEqual(newLocation)))
+                        {
+                            failedSourceLocations.Add(newLocation);
+                        }
+                    }
+                }
+            }
+
+            var expectedSourceLocations = expectedSourceLocationsStr
+                .Split("\n", StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries)
```
VeraBE, line 113:
```suggestion
                .Split(Environment.NewLine, StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries)
```

src/Analyzer.Core.UnitTests/templates/ParameterPassingFailModule.bicep
```diff
@@ -0,0 +1,23 @@
+param ftpsstate string
+param location string
+
+var sqlServerName = 'random'
```
VeraBE, line 4:
```suggestion
```

VeraBE, line 4: Can this line be removed?

src/Analyzer.Core/TemplateAnalyzer.cs
VeraBE: I don't understand why we don't dedup all of those results then, why do we want to return multiple identical results?

src/Analyzer.Types/Result.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// </summary>
+        public SourceLocation SourceLocation { get; }
+
+        /// <summary>
+        /// Gets or sets the JSON path to the location in the JSON where the rule was evaluated.
+        /// </summary>
+        public string JsonPath { get; set; }
+
+        /// <summary>
+        /// Gets the expression object associated with this result
```
anaismiller, line 29:
```suggestion
        /// Gets the expression object associated with this result.
```

src/Analyzer.BicepProcessor/BicepLocationResolver.cs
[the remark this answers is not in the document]
nonik0: yes I never hit this condition in all my testing and I removed it.

src/Analyzer.Core.UnitTests/TemplateAnalyzerTests.cs
```diff
@@ -41,9 +42,91 @@ public void AnalyzeTemplate_ValidInputValues_ReturnCorrectEvaluations(string res
             Assert.AreEqual(expectedEvaluationPassCount, evaluationsWithResults.Count(e => e.Passed));
         }
 
+        const string SimpleNestedFailExpectedSourceLocations = @"
```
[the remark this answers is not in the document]
nonik0: I actually reverted the changes here since I simplified the implementation for SourceLocation

src/Analyzer.Core.UnitTests/templates/ParameterPassingFailModule.bicep
nonik0: yes

src/Analyzer.BicepProcessor/Analyzer.BicepProcessor.csproj
```diff
@@ -9,8 +9,10 @@
   </ItemGroup>
   
   <ItemGroup>
-    <PackageReference Include="Azure.Bicep.Core" Version="0.8.9" />
+    <PackageReference Include="Azure.Bicep.Core" Version="0.11.1" />
+    <PackageReference Include="Azure.Bicep.Types" Version="0.3.28" /> <!-- temp -->
```
VeraBE, line 13: Can we already remove these temp references?

src/Analyzer.Core.UnitTests/templates/ParameterPassingFail.json
```diff
@@ -46,9 +46,6 @@
               "type": "string"
             }
           },
-          "variables": {
```
VeraBE, line 49: Was this removed by accident?

src/Analyzer.Reports/ConsoleReportWriter.cs
```diff
@@ -17,61 +16,59 @@ public class ConsoleReportWriter : IReportWriter
...
+                string fileMetadata;
+                fileMetadata = Environment.NewLine + Environment.NewLine + $"File: {fileWithResults}";
 
-            OutputToConsole(evaluations);
-        }
+                if (fileWithResults == templateFile.FullName)
+                {
+                    if (parametersFile != null)
+                    {
+                        fileMetadata += Environment.NewLine + $"Parameters File: {parametersFile}";
```
VeraBE, line 41: Oh, does this work okay after the changes on the PR to auto-discovery parameters files?

src/Analyzer.BicepProcessor/BicepLocationResolver.cs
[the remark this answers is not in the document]
nonik0: This was updated to be simplified.

src/Analyzer.BicepProcessor/Analyzer.BicepProcessor.csproj
nonik0: [New Bicep.Core](https://www.nuget.org/packages/Azure.Bicep.Core/0.12.1) released yesterday so should be able to!

src/Analyzer.Reports/ConsoleReportWriter.cs
nonik0: I did some ad hoc testing and it seemed to work but we also need to think through how we want to show results for files that may be evaluated twice, e.g. main.bicep refers to module.bicep, but there's also main.parameters.json. So, when scanning thatdirectory, module.bicep is going to be scanned twice--once as a module referenced by main.bicep with the parameters from main.parameters.json, and once as a standalone bicep file with no parameter file for itself.

How do we output this and have it not be confusing?? Or do we dedupe and maybe only show discrepancies?

src/Analyzer.Core.UnitTests/templates/ParameterPassingFail.json
nonik0: It's the analog to the comment you made on the bicep version of this file (in the module specifically). The bicep file was decompiled from this template and inherited the unnecessary variable.

src/Analyzer.Core/TemplateAnalyzer.cs
nonik0: So since the last comment, I've removed the "stack" from the source location so effectively a result will look identical for a rule whether it's found as part of a module or scanned in the entrypoint file. Also I tried removing this code assuming that the late deduping would cover it and interestingly TA-000021 fails so I need to investigate that further.

src/Analyzer.Core.UnitTests/templates/ParameterPassingFail.json
VeraBE: Oh, makes sense

src/Analyzer.Reports/ConsoleReportWriter.cs
VeraBE: mmm do we gain anything by scanning module.bicep by itself? Maybe we shouldn't even scan those files more than once?

src/Analyzer.Reports/ReportsHelper.cs
nonik0: OK--even with the new source data I hit a snag here. We need to be able to bucket each top-level evaluation by the source file it was evaluated in (specifically the resource that it evaluated against). However, I am seeing some evaluations that do not contain any results, and consequently no SourceLocations to pull the source file from. Using the TemplateIdentifier is inaccurate for results in modules. So, this means I don't know what file the evaluations are actually from if there's modules in the entrypoint.

Considering tabling this for a later PR considering it's an optimization and the cost is unknown but growing still.

src/Analyzer.Core/TemplateAnalyzer.cs
nonik0: The result deduping code now is in the Analyzer.Reports. As reporting is done outside/after TemplateAnalyze.AnalyzeTemplate() that means either we refactor to dedupe in TemplateAnalyzer or just update the tests to expect the dupes since they will be deduped in the reporting code (where we have tests specifically for deduping still).

VeraBE: I vote for the option where we don't need to expect duplicated results in the tests

src/Analyzer.BicepProcessor.UnitTests/BicepProcessorUnitTests.cs
```diff
@@ -0,0 +1,100 @@
+﻿using System;
```
anaismiller, line 1:
```suggestion
// Copyright (c) Microsoft Corporation.
// Licensed under the MIT License.

﻿using System;
```

src/Analyzer.BicepProcessor.UnitTests/BicepProcessorUnitTests.cs
```diff
@@ -0,0 +1,100 @@
...
+    {
+        [TestMethod]
+        public void ValidateBicepModuleMetadata()
+        {
+            string directory = Path.Combine(Directory.GetCurrentDirectory(), "templates");
+            string bicepPath = Path.Combine(directory, "TemplateWithMixedRefs.bicep");
+            (_, var bicepMetadata) = BicepTemplateProcessor.ConvertBicepToJson(bicepPath);
+            var actualModuleInfo = bicepMetadata.ModuleInfo.ToList();
+
+            // line numbers from Bicep.Core are 0-indexed (and maintained in module info)
```
anaismiller, line 23:
```suggestion
            // Line numbers from Bicep.Core are 0-indexed (and maintained in module info)
```

src/Analyzer.BicepProcessor.UnitTests/BicepProcessorUnitTests.cs
```diff
@@ -0,0 +1,100 @@
...
+using Microsoft.Azure.Templates.Analyzer.Types;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Analyzer.BicepProcessor.UnitTests
+{
+    [TestClass]
+    public class BicepProcessorUnitTests
+    {
+        [TestMethod]
+        public void ValidateBicepModuleMetadata()
```
anaismiller, line 16:
```suggestion
        public void ValidateBicepModuleMetadataCounts()
```
If I understood the assert?

src/Analyzer.BicepProcessor.UnitTests/BicepProcessorUnitTests.cs
```diff
@@ -0,0 +1,100 @@
...
+            var templateAnalyzer = TemplateAnalyzer.Create(false);
+            string directory = Path.Combine(Directory.GetCurrentDirectory(), "templates");
+            string armFilePath = Path.Combine(directory, "AppServicesLogs-Failures.json");
+            string armAsModuleFilePath = Path.Combine(directory, "TemplateWithTwoArmRefs.bicep");
+            string armTemplate = File.ReadAllText(armFilePath);
+
+            var armEvaluations = templateAnalyzer.AnalyzeTemplate(armTemplate, armFilePath);
+            var armAsModuleEvaluations = templateAnalyzer.AnalyzeTemplate(string.Empty, armAsModuleFilePath);
+
+            // verify only results in bicep fie are from ARM module
```
anaismiller, line 68:
```suggestion
            // Verify only results in bicep file are from ARM module
```

src/Analyzer.BicepProcessor.UnitTests/BicepProcessorUnitTests.cs
nonik0: The next line does the comparison of the actual modules so it validates more than counts. The combination of dict1.Count == dict2.Count && !(dict1.Except(dict2).Any()) is a shorthand for comparing two dictionaries.

src/Analyzer.Reports/ConsoleReportWriter.cs
```diff
@@ -17,61 +16,59 @@ public class ConsoleReportWriter : IReportWriter
...
+                    }
+                }
+                else
+                {
+                    fileMetadata += Environment.NewLine + $"Root Template: {templateFile}";
+                }
 
-        private void OutputToConsole(IEnumerable<Types.IEvaluation> evaluations)
```
anaismiller, line 34: Do we want to keep it as a separate func?

src/Analyzer.Reports/ConsoleReportWriter.cs
```diff
@@ -17,61 +16,59 @@ public class ConsoleReportWriter : IReportWriter
...
         public void WriteResults(IEnumerable<Types.IEvaluation> evaluations, IFileInfo templateFile, IFileInfo parametersFile = null)
         {
-            // Log info on file to be analyzed
-            string fileMetadata = Environment.NewLine + Environment.NewLine + $"File: {templateFile}";
-            if (parametersFile != null)
+            var resultsByFile = ReportsHelper.GetResultsByFile(evaluations, filesAlreadyOutput, out int passedEvaluations);
+
+            // output files in sorted order, but always output root first and regardless if no results
+            var filesWithResults = resultsByFile.Keys.ToList();
+            filesWithResults.Remove(templateFile.FullName);
```
anaismiller, line 28: Can there be duplicates in this list?

src/Analyzer.TemplateProcessor/Analyzer.TemplateProcessor.csproj
```diff
@@ -7,9 +7,9 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Azure.Deployments.Core" Version="1.0.635" />
-    <PackageReference Include="Azure.Deployments.Expression" Version="1.0.635" />
-    <PackageReference Include="Azure.Deployments.Templates" Version="1.0.635" />
+    <PackageReference Include="Azure.Deployments.Core" Version="1.0.694" />
```
anaismiller, line 10: Does the lastest version 1.0.765 also work here?

src/Analyzer.Types/TemplateContext.cs
```diff
@@ -27,9 +27,9 @@ public class TemplateContext
         public bool IsBicep { get; set; }
 
         /// <summary>
-        /// Source map that maps ARM JSON back to source Bicep
+        /// TODO Source map that maps ARM JSON back to source Bicep
```
anaismiller, line 30: is this tracked/blocking?

src/Analyzer.BicepProcessor.UnitTests/templates/AppServicesLogs-Failures.json
```diff
@@ -0,0 +1,33 @@
...
+      "apiVersion": "2019-08-01",
+      "type": "Microsoft.Web/sites",
+      "kind": "app",
+      "name": "diagLogsDisabledInSiteConfigProperty",
+      "location": "[parameters('location')]",
+      "properties": {
+        "siteConfig": {
+
+          "detailedErrorLoggingEnabled": false,
+
```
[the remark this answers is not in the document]
nonik0: I added a comment here but these are necessary for a test.

src/Analyzer.BicepProcessor/BicepTemplateProcessor.cs
```diff
@@ -73,7 +147,52 @@ public static (string, SourceMap) ConvertBicepToJson(string bicepPath)
...
+
+            // collect all needed module info from sourceFileGrouping metadata
+            var moduleInfo = sourceFileGrouping.UriResultByModule.Select(kvp =>
+            {
+                var bicepSourceFile = kvp.Key as BicepSourceFile;
+                var pathRelativeToEntryPoint = GetPathRelativeToEntryPoint(bicepSourceFile.FileUri.AbsolutePath);
+                var modules = kvp.Value.Values
+                    .Select(result =>
+                    {
+                        // do not include modules imported from public/private registries, as it is more useful for user to see line number
```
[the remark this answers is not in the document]
nonik0: This should not create parsing issues. This module data is part of bicep metadata that's only used post-analysis. Modules in registries can cause bicep compilation errors if not available/accessible, but that's expected and captured already in errors.

src/Analyzer.Reports/ConsoleReportWriter.cs
nonik0: The list is from the keys of a dictionary so it should not have duplicates.

src/Analyzer.BicepProcessor.UnitTests/templates/AppServicesLogs-Failures.bicep
```diff
@@ -0,0 +1,18 @@
...
+
+resource diagLogsDisabledInSiteConfigProperty 'Microsoft.Web/sites@2019-08-01' = {
+  kind: 'app'
+  name: 'diagLogsDisabledInSiteConfigProperty'
+  location: location
+  properties: {
+    siteConfig: {
+
+      detailedErrorLoggingEnabled: false
+
```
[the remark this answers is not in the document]
nonik0: removed from bicep file as not necessary like ARM template

src/Analyzer.Types/TemplateContext.cs
nonik0: Nope just need to update this as it's now bicep metadata (source map and module info). Good catch.

src/Analyzer.TemplateProcessor/Analyzer.TemplateProcessor.csproj
nonik0: 1.0.750 is latest for me

src/Analyzer.Reports/ConsoleReportWriter.cs
nonik0: The implementation changed and preserving the separate function wasn't as clean or made sense as much to do. Let me know if you have a way to improve readability.

src/Analyzer.Core/TemplateAnalyzer.cs
VeraBE: Especially because this also affects the built-in rule tests, it adds so much noise to repeat the description of each failed line number for each repeated result:
https://github.com/Azure/template-analyzer/blob/87e6e2bb8e9228521c855c857df2cdb257545a68/src/Analyzer.Core.BuiltInRuleTests/Tests/TA-000021/TA-000021.json

I believe it was one of the reasons for this issue: https://github.com/Azure/template-analyzer/issues/228

If we are closing it without fixing this I think we should at least open a new one

src/Analyzer.Reports.UnitTests/SarifReportWriterE2ETests.cs
```diff
@@ -116,39 +139,68 @@ public void AnalyzeTemplateTests()
         }
 
         [DataTestMethod]
-        [DataRow(true, DisplayName = "Templates in root directory")]
-        [DataRow(true, "subPath", DisplayName = "Template in subdirectory")]
-        [DataRow(false, "..", "anotherRepo", "subfolder", DisplayName = "Templates under root directory and outside of root")]
-        public void AnalyzeDirectoryTests(bool secondTemplateUsesRelativePath, params string[] secondTemplatePathPieces)
+        //[DataRow("RedisCache.json", "SQLServerAuditingSettings.json", null, DisplayName = "Templates in root directory")]
```
VeraBE, line 142: Are these lines intentionally commented out?

src/Analyzer.Core/TemplateAnalyzer.cs
nonik0: That's interesting because I read this quote from #228: "_Reports_ could then do something simple like the following..." and concluded that the reports were the place to do the deduping. If we do want to do deduping in TemplateAnalyzer class, we will need deduping in the reports as we still need to dedupe across files with modules. So, I guess I am not sure what the original intent of where the deduping lives, but I am open to your notion that there's not dupes from TemplateAnalyzer itself. I think we could make use of distinct() here to clean up the temp deduping code I removed as one option.

VeraBE: Okay, now I'm not so sure what was the original idea of the issue, but when I linked PR #262 to it I was referring to the duplications in the built-in rule tests

I'm not sure if I understood the last part either, do you mean adding a .Distinct() in the TestRunner to avoid repeated descriptions in the test files? That sounds good to me

src/Analyzer.Reports/ReportsHelper.cs
```diff
@@ -0,0 +1,121 @@
...
+
+            return (actualFile, failedResults);
+        }
+
+        /// <summary>
+        /// Gets all failed results in an evaluation and returns in a flattened list
+        /// </summary>
+        /// <param name="evaluation">The evaluation to get results from</param>
+        /// <param name="failedResults">Accumulator used inrecursive calls</param>
+        /// <returns></returns>
```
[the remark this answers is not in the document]
VeraBE: @nonik0 did these comments got lost in the "hidden items" section?

src/Analyzer.Core/TemplateAnalyzer.cs
nonik0: OK I see what you mean, I don't think that's what meant with your suggestion, but I like it better anyway to clean up the duplication in the built in rule tests. Will do this shortly.

src/Analyzer.BicepProcessor/BicepLocationResolver.cs
```diff
@@ -34,20 +43,69 @@ public BicepLocationResolver(TemplateContext templateContext)
...
+            }
+
+            // check if match is an ARM module reference, if so return location in that template
+            var moduleMetadata = this.metadata.ModuleInfo.FirstOrDefault(info => info.FileName == bestMatch.FilePath);
+            if (moduleMetadata != default && moduleMetadata.Modules.ContainsKey(bestMatch.SourceLine.Value))
+            {
+                var modulePath = moduleMetadata.Modules[bestMatch.SourceLine.Value];
+                if (modulePath.EndsWith(".json") && File.Exists(modulePath))
+                {
+                    var template = ArmTemplateCache.GetArmTemplate(modulePath);
```
nonik0, line 78: This solution for getting line number does not work for nested ARM templates. See #307.

src/Analyzer.Reports/ReportsHelper.cs
nonik0: #308 to track. Not sure we have an issue yet specifically with the evaluations with no results. I want to investigate whether this only happens with passing evaluations, as in that case we could duplicate the evaluation as it would apply to each dependent template.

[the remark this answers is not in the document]
nonik0: No this is actual to prevent output results for the same file twice if it has multiple references. Without this, the test case "RedisCache template results have multiple references, file results should be deduped" will fail, as the module file will have results for itself as well as its parent file. This is unnecessary after #308 is fixed.

src/Analyzer.BicepProcessor/BicepTemplateProcessor.cs
```diff
@@ -2,67 +2,141 @@
...
 using Bicep.Core.Workspaces;
+using Microsoft.Azure.Templates.Analyzer.Types;
+using Microsoft.WindowsAzure.ResourceStack.Common.Extensions;
 
 namespace Microsoft.Azure.Templates.Analyzer.BicepProcessor
 {
+    /// <summary>
+    /// Helper class containing metadata from Bicep compilation necessary for later analysis
+    /// </summary>
+    public class BicepMetadata
```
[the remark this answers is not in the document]
nonik0: Good call, I moved the new metadata classes to another file.

src/Analyzer.BicepProcessor/BicepLocationResolver.cs
```diff
@@ -34,20 +43,69 @@ public BicepLocationResolver(TemplateContext templateContext)
...
+                    }
+                }
+            }
+
+            var entrypointFullPath = Path.GetDirectoryName(this.EntrypointFilePath);
+            var matchFullFilePath = Path.GetFullPath(Path.Combine(entrypointFullPath, bestMatch.FilePath));
+            return new SourceLocation(matchFullFilePath, bestMatch.SourceLine.Value + 1); // convert line number back to 1-indexing
+        }
+
+        private static class ArmTemplateCache
```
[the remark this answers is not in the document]
nonik0: At this point it's small enough where I am ok with leaving it as is. This code will be changing soon regardless due to #307.

VeraBE (approved):

src/Analyzer.BicepProcessor/BicepLocationResolver.cs
```diff
@@ -34,20 +43,69 @@ public BicepLocationResolver(TemplateContext templateContext)
...
+            var bestMatch = metadata.SourceMap.Entries
+                .Select(sourceFile =>
+                {
+                    var match = sourceFile.SourceMap.FirstOrDefault(mapping => mapping.TargetLine == jsonLine);
+                    var matchSize = match != default
+                        ? sourceFile.SourceMap.Count(mapping => mapping.SourceLine == match.SourceLine)
+                        : int.MaxValue;
+                    return (sourceFile.FilePath, match?.SourceLine, matchSize);
+                })
+                .MinBy(tuple => tuple.matchSize);
```
anaismiller, line 63: Can you explain why you are doing the min here to be the best match by updating it in the // Comment

src/Analyzer.BicepProcessor/BicepLocationResolver.cs
```diff
@@ -34,20 +43,69 @@ public BicepLocationResolver(TemplateContext templateContext)
...
+            if (moduleMetadata != default && moduleMetadata.Modules.ContainsKey(bestMatch.SourceLine.Value))
+            {
+                var modulePath = moduleMetadata.Modules[bestMatch.SourceLine.Value];
+                if (modulePath.EndsWith(".json") && File.Exists(modulePath))
+                {
+                    var template = ArmTemplateCache.GetArmTemplate(modulePath);
+                    var token = template.InsensitiveToken(pathInExpandedTemplate, InsensitivePathNotFoundBehavior.LastValid);
+                    var lineNumber = (token as IJsonLineInfo)?.LineNumber;
+
+                    // fall back to location of module refernce in parent file if failing to get line number
```
anaismiller, line 82:
```suggestion
                    // Fall back to location of module reference in parent file if failing to get line number
```

src/Analyzer.BicepProcessor/BicepLocationResolver.cs
```diff
@@ -34,20 +43,69 @@ public BicepLocationResolver(TemplateContext templateContext)
...
+
+        private static class ArmTemplateCache
+        {
+            private static readonly Dictionary<string, JObject> Templates = new();
+
+            public static JObject GetArmTemplate(string templatePath)
+            {
+                if (!Templates.ContainsKey(templatePath))
+                {
+                    // assumption: template has already been previously parsed by Bicep library and was valid
```
anaismiller, line 103:
```suggestion
                    // Assumption: template has already been parsed by Bicep library and is valid
```

src/Analyzer.BicepProcessor/BicepTemplateProcessor.cs
```diff
@@ -26,43 +31,50 @@ namespace Microsoft.Azure.Templates.Analyzer.BicepProcessor
...
+                    new ContainerRegistryClientFactory(new TokenCredentialFactory()),
+                    new TemplateSpecRepositoryFactory(new TokenCredentialFactory()),
+                    featureProviderFactory,
+                    configurationManager),
+                configurationManager);
             var workspace = new Workspace();
-            var sourceFileGrouping = SourceFileGroupingBuilder.Build(fileResolver, moduleDispatcher, workspace, PathHelper.FilePathToFileUrl(bicepPath), configuration);
+            var sourceFileGrouping = SourceFileGroupingBuilder.Build(fileResolver, moduleDispatcher, workspace, PathHelper.FilePathToFileUrl(bicepPath));
 
             // pull modules optimistically
```
anaismiller, line 69:
```suggestion
            // Pull modules optimistically
```

src/Analyzer.BicepProcessor/BicepTemplateProcessor.cs
```diff
@@ -26,43 +31,50 @@ namespace Microsoft.Azure.Templates.Analyzer.BicepProcessor
...
-            if (moduleDispatcher.RestoreModules(configuration, moduleDispatcher.GetValidModuleReferences(sourceFileGrouping.ModulesToRestore, configuration)).Result)
+            if (moduleDispatcher.RestoreModules(moduleDispatcher.GetValidModuleReferences(sourceFileGrouping.GetModulesToRestore())).Result)
             {
                 // modules had to be restored - recompile
```
anaismiller, line 72:
```suggestion
                // Modules had to be restored - recompile
```

src/Analyzer.BicepProcessor/BicepTemplateProcessor.cs
```diff
@@ -73,7 +147,52 @@ public static (string, SourceMap) ConvertBicepToJson(string bicepPath)
                 throw new Exception($"Bicep issues found:{Environment.NewLine}{string.Join(Environment.NewLine, bicepIssues)}");
             }
 
-            return (stringWriter.ToString(), emitResult.SourceMap);
+
```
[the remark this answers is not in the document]
anaismiller: still there

src/Analyzer.BicepProcessor/BicepTemplateProcessor.cs
```diff
@@ -73,7 +85,52 @@ public static (string, SourceMap) ConvertBicepToJson(string bicepPath)
                 throw new Exception($"Bicep issues found:{Environment.NewLine}{string.Join(Environment.NewLine, bicepIssues)}");
             }
 
-            return (stringWriter.ToString(), emitResult.SourceMap);
+            
+
+            string GetPathRelativeToEntryPoint(string absolutePath) => Path.GetRelativePath(
+                Path.GetDirectoryName(sourceFileGrouping.EntryPoint.FileUri.AbsolutePath), absolutePath);
+
+            // collect all needed module info from sourceFileGrouping metadata
```
anaismiller, line 93:
```suggestion
            // Collect all needed module info from sourceFileGrouping metadata
```

src/Analyzer.BicepProcessor/BicepTemplateProcessor.cs
```diff
@@ -73,7 +85,52 @@ public static (string, SourceMap) ConvertBicepToJson(string bicepPath)
...
+            var moduleInfo = sourceFileGrouping.UriResultByModule.Select(kvp =>
+            {
+                var bicepSourceFile = kvp.Key as BicepSourceFile;
+                var pathRelativeToEntryPoint = GetPathRelativeToEntryPoint(bicepSourceFile.FileUri.AbsolutePath);
+                var modules = kvp.Value.Values
+                    .Select(result =>
+                    {
+                        // do not include modules imported from public/private registries, as it is more useful for user to see line number
```
anaismiller, line 101:
```suggestion
                        // Do not include modules imported from public/private registries, as it is more useful for user to see line number
```

src/Analyzer.BicepProcessor/BicepTemplateProcessor.cs
```diff
@@ -73,7 +85,52 @@ public static (string, SourceMap) ConvertBicepToJson(string bicepPath)
...
+                            || moduleDeclaration.Path is not StringSyntax moduleDeclarationPath
+                            || moduleDeclarationPath.SegmentValues.Any(v => IsModuleRegistryPathRegex.IsMatch(v)))
+                        {
+                            return null;
+                        }
+
+                        var moduleLine = TextCoordinateConverter.GetPosition(bicepSourceFile.LineStarts, moduleDeclaration.Span.Position).line;
+                        var modulePath = result.FileUri.AbsolutePath;
+
+                        // use relative paths for bicep to match file paths used in bicep modules and source map
```
anaismiller, line 113:
```suggestion
                        // Use relative paths for bicep to match file paths used in bicep modules and source map
```

src/Analyzer.Reports/ReportsHelper.cs
```diff
@@ -0,0 +1,91 @@
+﻿using System;
```
anaismiller, line 1: add license line

src/Analyzer.Reports/ReportsHelper.cs
```diff
@@ -0,0 +1,91 @@
...
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using Microsoft.Azure.Templates.Analyzer.Types;
+
+namespace Microsoft.Azure.Templates.Analyzer.Reports
+{
+    /// <summary>
+    /// Helper class for ReportsWriters
```
anaismiller, line 10:
```suggestion
    /// Helper class for ReportsWriter
```

src/Analyzer.Reports/ReportsHelper.cs
```diff
@@ -0,0 +1,91 @@
...
+            var resultsByFile = new Dictionary<string, List<(IEvaluation, IList<Result>)>>();
+            passedEvaluations = 0;
+
+            foreach (var evaluation in evaluations)
+            {
+                if (!evaluation.Passed)
+                {
+                    (var actualFile, var failedResults) = GetResultsByFileInternal(evaluation);
+
+                    // a file's results may have already been output if analyzing directory
```
anaismiller, line 48:
```suggestion
                    // A file's results may have already been output if analyzing directory
```

src/Analyzer.Reports/ReportsHelper.cs
```diff
@@ -0,0 +1,91 @@
...
+                    {
+                        continue;
+                    }
+
+                    if (!resultsByFile.ContainsKey(actualFile))
+                    {
+                        resultsByFile[actualFile] = new List<(IEvaluation, IList<Result>)>();
+                    }
+
+                    // skip any evaluations with duplicate results (i.e. two source locations from other templates refer to same result)
```
anaismiller, line 59:
```suggestion
                    // Skip any evaluations with duplicate results (i.e. two source locations from other templates refer to same result)
```

src/Analyzer.Reports/ReportsHelper.cs
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        /// <summary>
+        /// Gets all failed results in an evaluation, sorts, and returns rule ID
+        /// </summary>
+        /// <param name="evaluation">The evaluation to get results for</param>
+        /// <returns>A list of distinct failed results</returns>
+        public static (string, IList<Result>) GetResultsByFileInternal(Types.IEvaluation evaluation)
+        {
+            // get all distinct failed results in evaluation
```
anaismiller, line 82:
```suggestion
            // Get all distinct failed results in evaluation
```

src/Analyzer.Reports.UnitTests/SarifReportWriterE2ETests.cs
```diff
@@ -84,6 +89,24 @@ public void AnalyzeTemplateTests()
                 expectedLinesForRun.ContainsKey(result.RuleId).Should().BeTrue("Unexpected result found in SARIF");
                 result.Level.Should().Be(FailureLevel.Error);
 
+                // determine which template file was evaluated for this SARIF result (all locations will be in same file, so taking first)
```
anaismiller, line 92:
```suggestion
                // Determine which template file was evaluated for this SARIF result (all locations will be in same file, so taking first)
```

src/Analyzer.Reports.UnitTests/SarifReportWriterE2ETests.cs
```diff
@@ -201,45 +253,57 @@ public void AnalyzeDirectoryTests(bool secondTemplateUsesRelativePath, params st
...
-                    file: "RedisCache.json",
+                    file: firstTemplate,
                     uriBase: SarifReportWriter.UriBaseIdString,
                     lines: new List<List<int>> {
                         new List<int> { 1 }
                     })
                 }
             };
 
+            // extra validation for extra results in parent bicep template
```
anaismiller, line 295:
```suggestion
            // Extra validation for extra results in parent bicep template
```

src/Analyzer.BicepProcessor/BicepTemplateProcessor.cs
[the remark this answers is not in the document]
nonik0: Sorry, the code snippet/line number shown here threw me off and I thought I had already gotten it. I found some extra lines though and removed them so should be OK now.

anaismiller (approved):

PR #399, opened 2025-08-12
Support for Language Version 2 in ARM templates

### LINKED ISSUE #388: [BUG] Exception "Cannot cast Newtonsoft.Json.Linq.JProperty to Newtonsoft.Json.Linq.JToken." when analyzing template using AVM module
### Describe the bug

We've got some bicep that makes use of the container-registry AVM module and when running it through the template analyzer (using the AzDO security task) we get the following exception "Cannot cast Newtonsoft.Json.Linq.JProperty to Newtonsoft.Json.Linq.JToken". I'm not sure what part of the AVM module is causing this and I don't know enough about how this tool works to properly debug what might be going on here.

This happens whether I try to analyze the bicep file or the built json file.

### Expected behavior

Bicep is analyzed and reports back an errors/warnings without throwing an exception.

### Reproduction Steps

I've been able to get the exception when using just this as the bicep:

```
module acr 'br/public:avm/res/container-registry/registry:0.7.0' = {
  name: 'testing'
  params: {
    name: 'testing'
  }
}
```

### Environment

TemplateAnalyzer version 0.8.0 when run in AzDO or latest of main when run locally
.net version is 8.0.405
IDE is VS Code 1.96.4

### LINKED ISSUE #390: [FEATURE REQ] Support ARM template languageVersion 2.0

ARM templates (in JSON) now have a [languageVersion 2.0](https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/syntax#languageversion-20) enhancement. If this top-level property is specified on the template, several enhanced features are available in the schema language.

Importantly, this also changes the `"resources"` property from an array to an object, where each resource is a named property instead of an array element.

### Additional context

_No response_

### LINKED ISSUE #391: [BUG] Template Analyzer doesn't understand languageVersion 2.0
### Describe the bug

While #390 tracks supporting these kinds of templates for analysis, if this kind of template is encountered currently, some new features break some assumptions in template parsing, and scanning fails; specifically, the assumption that `"resources"` is an array is definitely broken.

### Expected behavior

Ideally, these templates should be supported and scanned appropriately. At the very least, these should be detected and gracefully fail (with an understandable message) instead of throwing an exception and failing.

### Reproduction Steps

Scan any template that uses languageVersion 2.0 which has at least one resource.

### Environment

_No response_

JohnathonMohr:
## Description

Support new features introduced in [language version 2](https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/syntax#languageversion-20). (Closes #388, #390, #391)

- Support resources with symbolic names in addition to array format
- Improvements to nested template discovery
- Update Bicep and Deployments core packages
- Fix VS Code launch profiles

---


src/Analyzer.Core/TemplateAnalyzer.cs
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft Corporation.
+﻿// Copyright (c) Microsoft Corporation.
```
reynoldskr, line 1: What did you do here?

src/Analyzer.Core/TemplateAnalyzer.cs
```diff
@@ -178,62 +179,58 @@ private IEnumerable<IEvaluation> AnalyzeAllIncludedTemplates(string populatedTem
...
-                            if (parametersToPass != null)
-                            {
-                                parametersToPass["parameters"] = parametersToPass;
-                                parametersToPass["$schema"] = "https://schema.management.azure.com/schemas/2019-04-01/deploymentParameters.json#";
-                                parametersToPass["contentVersion"] = "1.0.0.0";
-                                parameters = JsonConvert.SerializeObject(parametersToPass);
-                            }
+                        if (parametersToPass != null)
+                        {
+                            parametersToPass["parameters"] = parametersToPass;
```
reynoldskr, line 223: Just checking, we don't need to specify lang version 2 here?

src/Analyzer.Reports.NuGet/Analyzer.Reports.nuspec
```diff
@@ -14,8 +14,8 @@
             <group targetFramework="net8.0">
                 <dependency id="Newtonsoft.json" version="13.0.3" />
                 <dependency id="Sarif.Sdk" version="2.4.16" />
-                <dependency id="System.IO.Abstractions" version="20.0.4" />
-                <dependency id="Microsoft.Extensions.Logging.Abstractions" version="6.0.1" />
+                <dependency id="System.IO.Abstractions" version="21.3.1" />
```
reynoldskr, line 17: What would you think of migrating from separate nuspecs here to put this inside the csproj, since we're all SDK-style/.net core?

reynoldskr: Not necessarily *in this PR* just in general

reynoldskr (approved):
Nothing blocking

src/Analyzer.Core/TemplateAnalyzer.cs
JohnathonMohr: Nothing, this line doesn't appear changed in my local commits. I've seen GH do this before, but I don't know what it thinks it's picking up.

src/Analyzer.Reports.NuGet/Analyzer.Reports.nuspec
JohnathonMohr: I would ultimately prefer that, but the csproj file doesn't support our package structure - we only publish a few packages (not one-per-csproj) and last I checked this isn't supported with a csproj specification.

src/Analyzer.Core/TemplateAnalyzer.cs
JohnathonMohr: As far as I've seen, the language version spec only applies to a template, not the parameters. So we don't need to specify it in the parameter file.

src/Analyzer.Utilities.UnitTests/JsonPathResolverTests.cs
```diff
@@ -104,19 +101,15 @@ public void Resolve_JsonContainsPath_ReturnsResolverWithCorrectJtokenAndPath(str
                     },
                 }");
 
-            var resolver = new JsonPathResolver(jtoken, jtoken.Path);
+            var resolver = new JsonPathResolver(jobject, jobject.Path);
 
-            // Do twice to verify internal cache correctness
```
anaismiller, line 109: should this checking still be here ("// Do twice to verify internal cache correctness")?

anaismiller: and likewise is other spots in this file?

src/Analyzer.Utilities.UnitTests/JsonPathResolverTests.cs
```diff
@@ -255,6 +235,174 @@ public void ResolveResourceType_JObjectWithResourcesNotArrayOfObjects_ReturnsEmp
...
+
+            // Validate that the results from the first and second resolution of each path are the same.
+            for (int i = 0; i < pathsToResolve.Length; i++)
+            {
+                Assert.AreEqual(1, resolvedResults[0, i].Count);
+                Assert.AreEqual(1, resolvedResults[1, i].Count);
+                Assert.AreEqual<object>(resolvedResults[0, i][0].JToken, resolvedResults[1, i][0].JToken);
+            }
+        }
+
```
anaismiller, line 405: Are there cases where both lang v1 and v2 can be declared in the same file? Would it cause issues?
What about in the same repo with a mix of nested templates in both languages?

src/Analyzer.Utilities.UnitTests/JsonPathResolverTests.cs
```diff
@@ -255,6 +235,174 @@ public void ResolveResourceType_JObjectWithResourcesNotArrayOfObjects_ReturnsEmp
             Assert.AreEqual(0, new JsonPathResolver(jToken, jToken.Path).ResolveResourceType("anything").Count());
         }
 
+        [TestMethod]
+        public void ResolveResourceType_LanguageVersion2SymbolicNaming_ReturnsResourcesOfCorrectType()
```
anaismiller, line 239: To be very explicit, can you update a test (if one exists)/add a new test for `ResolveResourceType_LanguageVersion1SymbolicNaming_ReturnsResourcesOfCorrectType()`? It could help catch errors if someone makes updates to see if it was a lang version error

anaismiller (approved):
I remember last time, there was a file you missed updating for the release. Just verifying, did you update that file? (for a future enhancement, whatever that file was, you can add it to the PR checklist)

src/Analyzer.Utilities.UnitTests/JsonPathResolverTests.cs
JohnathonMohr: I removed these loops for two reasons:
- Testing the same core functionality in multiple tests isn't good practice - it should be done in a single test
- These weren't great tests anyway - although it verified the second lookup was the same as the first, the test could still pass if, for example, the cache functionality was removed entirely (at which point it's just pointless additional work in the test process).

As an improvement, I removed all these loops and instead created a new, dedicated test doing only cache validation (and it does a much more thorough job at validating the functionality).

anaismiller: Sounds like a great improvement, thank you for explaining!

src/Analyzer.Utilities.UnitTests/JsonPathResolverTests.cs
JohnathonMohr: The most important pieces for lang v2 are symbolic names and the object structure of 'resources' (vs array). That being said:
- I don't know if a file can be both lang 1 and 2 (it could only change within a nested template). But since resource mapping and type resolution supports both, it should be opaque.
- Multiple types in the same repo is definitely possible, but they will be analyzed separately anyway, since this doesn't support linked templates right now.

src/Analyzer.Utilities.UnitTests/JsonPathResolverTests.cs
JohnathonMohr: Can you clarify? version 1 doesn't support symbolic naming, so I don't see the value in testing it.
Additionally:
- `JsonPathResolver` doesn't actually reference the version, it just checks for both styles
- A template attempting to use symbolic names with v1 won't deploy anyway, and I want to avoid template correctness validation.

anaismiller: You answered my concern!

